PR #7, opened 2018-05-23
Added possibility to query all fields from a specific table

jimmya:
Referenced from: https://github.com/vapor/fluent/issues/471

Hi,

I'm trying to build the following query using Fluent:
```
SELECT
	events.*
FROM events
	JOIN followers ON events."userID" = followers."userID" 
WHERE followers."followerID" = 2
```
I only want all values from `event` returned. In order to achieve this I'm using the `.customSQL` function. There is however a limiting factor. The way it is currently serialized the `name` of a column is always escaped. Resulting in the following query (postgreSQL):
```
SELECT
	"events"."*"
```
which of course would lead to errors as `*` is not a column of `events`.

My suggestion is to modify `DataColumn.name` to enable either requesting all (`*`) or a custom column specified by name (which would be escaped).

Edit:

I ended up adding a case to `DataQueryColumn` as this resulted in the least impact and cleaner syntax. And I believe a non breaking change but correct me if I'm wrong.

### DIFF
--- a/Sources/SQL/DataQuery/DataQueryColumn.swift
+++ b/Sources/SQL/DataQuery/DataQueryColumn.swift
@@ -2,6 +2,9 @@
 public enum DataQueryColumn {
     /// All columns, `*`.
     case all
+    
+    /// All columns of a table, `foo`.*
+    case tableAll(table: String)
 
     /// A single `DataColumn` with optional key.
     case column(DataColumn, key: String?)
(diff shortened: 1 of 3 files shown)


jimmya:
Another possible solution could be by modifying the `all` enum case.
Something like this:
```
case all(table: String?)
```
Where a nil value for `table` would result in `*` and when entering a table name `tablename.*`. This would be a breaking change however.

jimmya:
I'm able to reproduce the mentioned query in the opening post by executing the following fluent sql query if this change where to be implemented:
```
Event.query(on: req)
  .join(field: \Follower.userID, to: \Event.userID)
  .filter(Follower.self, \Follower.followerID == 2)
  .customSQL { (query) in
      if case .query(var dataQuery) = query {
          dataQuery.columns = [DataQueryColumn.tableAll(table: Event.entity)]
          query = .query(dataQuery)
      }
  }
  .all()
```

penny-coin:
Hey @jimmya, you just merged a pull request, have a coin! 

You now have 1 coins.

PR #49, opened 2019-01-11
Add support for NULL/NOT NULL-based sorting

Cellane:
This short PR adds support for `ORDER BY IS NULL`/`ORDER BY IS NOT NULL` clauses which causes rows with/without value present to be prioritized over other rows.


penny-coin:
Hey @Cellane, you just merged a pull request, have a coin! 

You now have 142 coins.

PR #61, opened 2019-07-02
Use standard SQL 92 operator for not equal.

johan-carlberg:
Not every database supports the proprietary operator "!=" as a not equal operator.


kerusan:
@tanner0101 Why is this pull request not approved?

kerusan:
Thanx!

PR #91, opened 2020-02-03
Update SQLBinaryOperator.swift to include "like" and "notLike"

proggeramlug:
Added `like` and `notLike` to the `serialize`

I don't see why they would not be added? It throws me an error without it when trying to express "LIKE" in MySQL. This might not be the only fix needed but I figured they can and should be included anyways.


proggeramlug:
@calebkleveter check this out

gwynne:
I believe the concern was less with the `LIKE` operator itself and more with the specific "likeness" syntax supported by different databases. This would be a matter for the `SQLDialect` implementations in each driver, though, and the operator itself is unaffected.

PR #102, opened 2020-04-12
Adds a `normalizeSQLConstraint` method

mcdappdev:
Adds a `normalizeSQLConstraint` method to `SQLDialect` so that the drivers can truncate constraint identifiers that are too long.

### DIFF
--- a/Sources/SQLKit/Query/SQLConstraint.swift
+++ b/Sources/SQLKit/Query/SQLConstraint.swift
@@ -38,15 +38,25 @@ extension SQLDialect {
         let utf8Midpoint = identifier.index(identifier.startIndex, offsetBy: midPoint).samePosition(in: identifier.utf8)!
         let excessInBytes = identifier.utf8.count - self.maximumConstraintIdentifierLength // number of *bytes* by which the string is too long
         let excessCutdown = excessInBytes >> 1 // number of bytes on either side of the midpoint to remove
-        let utf8PreCutdownIndex = identifier.utf8.index(utf8Midpoint, offsetBy: -excessCutdown)
-        let utf8PostCutdownIndex = identifier.utf8.index(utf8Midpoint, offsetBy: excessCutdown)
-        let realPreCutdownIndex = utf8PreCutdownIndex.samePosition(in: identifier)!
-        let realPostCutdownIndex = utf8PostCutdownIndex.samePosition(in: identifier)!
-        let cutdownRange = realPreCutdownIndex...realPostCutdownIndex
+        var utf8PreCutdownIndex = identifier.utf8.index(utf8Midpoint, offsetBy: -excessCutdown)
+        var realPreCutdownIndex = utf8PreCutdownIndex.samePosition(in: identifier)
+        while realPreCutdownIndex == nil && utf8PreCutdownIndex > identifier.utf8.startIndex {
+            identifier.utf8.formIndex(before: &utf8PreCutdownIndex)
+            realPreCutdownIndex = utf8PreCutdownIndex.samePosition(in: identifier)
+        }
+        
+        var utf8PostCutdownIndex = identifier.utf8.index(utf8Midpoint, offsetBy: excessCutdown)
+        var realPostCutdownIndex = utf8PreCutdownIndex.samePosition(in: identifier)
+        while realPostCutdownIndex == nil && utf8PreCutdownIndex > identifier.utf8.endIndex {
+            identifier.utf8.formIndex(after: &utf8PostCutdownIndex)
+            realPostCutdownIndex = utf8PreCutdownIndex.samePosition(in: identifier)
+        }
+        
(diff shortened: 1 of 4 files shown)


Sources/SQLKit/Query/SQLConstraint.swift
```diff
@@ -17,9 +17,23 @@ public struct SQLConstraint: SQLExpression {
...
             serializer.write(" ")
         }
         self.algorithm.serialize(to: &serializer)
     }
 }
+
+extension SQLDialect {
+    public func normalizeSQLConstraintIdentifier(_ identifier: String) -> String {
+        guard identifier.utf8.count > self.maximumConstraintIdentifierLength else { return identifier }
+        let midPoint = (identifier.count >> 1) - ((identifier.utf8.count - self.maximumConstraintIdentifierLength) >> 1)
```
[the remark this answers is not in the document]
gwynne: Isn't working with strings "correctly" just a ton of fun?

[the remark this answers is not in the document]
gwynne: Oh, and add tests which include complex Unicode characters in the test strings, such as `"smileyIdentifier😀IsASmiley"` (simple non-BMP emoji), `"ohWellWhatA🤷🏻‍♀️KindOfDay"` (complex variation selector emoji), and `"checkOutHowĈôm̂b̂în̂în̂ĝT⃠h⃠i⃠n⃠g⃠s⃠Works"` (combining characters on code points, which is multiple codepoints but only one grapheme cluster).

gwynne:
Note: None of the drivers were updated to provide their own values for the `maximumConstraintIdentifierLength` property on `SQLDialect` because:

- **Postgres**: Limit is `63`, same as the default (since the algorithm conservatively assumes length - 1)
- **MySQL**: Limit is `64`, same as the default
- **SQLite**: Has no limit, so have it use the default to prevent surprising behavior for those switching drivers after starting with SQLite.
- **Mongo**: Not an SQL database anyway.

Sources/SQLKit/Query/SQLConstraint.swift
```diff
@@ -17,9 +17,27 @@ public struct SQLConstraint: SQLExpression {
...
+                name.serialize(to: &serializer)
+            }
             serializer.write(" ")
         }
         self.algorithm.serialize(to: &serializer)
     }
 }
+
+extension SQLDialect {
+    public func normalizeSQLConstraintIdentifier(_ identifier: String) -> String {
```
tanner0101, line 33: Should this be a protocol requirement with a default implementation so that the dialect can choose what it wants to do?

Tests/SQLKitTests/SQLKitTests.swift
```diff
@@ -259,6 +259,17 @@ CREATE TABLE `planets`(`id` BIGINT PRIMARY KEY AUTOINCREMENT, `name` TEXT DEFAUL
 """
                        )
     }
+    
+    func testConstraintLengthNormalization() {
+        XCTAssertEqual(
+            db.dialect.normalizeSQLConstraintIdentifier("fk:obnoxiously_long_table_name.other_table_name_id+other_table_name.id"),
```
tanner0101, line 265: Maybe we should put a `...` in the middle so the truncation is clearer? 

```
// too long
fk:foo.bar+baz.quux
// truncates to
fk:fo...ux
```

Sources/SQLKit/Query/SQLConstraint.swift
```diff
@@ -17,7 +17,12 @@ public struct SQLConstraint: SQLExpression {
     public func serialize(to serializer: inout SQLSerializer) {
         if let name = self.name {
             serializer.write("CONSTRAINT ")
-            name.serialize(to: &serializer)
+            if let identifier = (name as? SQLIdentifier)?.string {
```
tanner0101, line 20: Casting as `SQLIdentifier` here could cause problems.

Sources/SQLKit/SQLDialect.swift
```diff
@@ -14,6 +14,7 @@ public protocol SQLDialect {
     var triggerSyntax: SQLTriggerSyntax { get }
     var alterTableSyntax: SQLAlterTableSyntax { get }
     func customDataType(for dataType: SQLDataType) -> SQLExpression?
+    func normalizeSQLConstraintIdentifier(identifier: String) -> String
```
tanner0101, line 17: This should accept and return `SQLExpression`.

tanner0101: `identifier` is mentioned twice here. I'd use either:
```swift
normalizeSQLConstraintIdentifier(_:)
// or
normalizeSQLConstraint(identifier:)
```

tanner0101:
These changes are now available in [3.0.0-rc.2](https://github.com/vapor/sql-kit/releases/tag/3.0.0-rc.2)

PR #109, opened 2020-06-18
Let's make the CI even more radical

gwynne:


MrLotU:
Would be awesome if you could tell GH to only do master builds if it didn't already do a PR build for that commit, but anyhow

gwynne:
I can't even make the blasted thing handle macOS sanely... 😭

PR #111, opened 2020-06-22
Add ways to construct SQLQueryString

t-ae:
Currently `SQLQueryString` can be constructed only through string literals and interpolations.
But we often want to build it conditionally.
This PR adds 3 functions to help constructing it:

- `SQLQueryString.stringInterpolation(SQLQueryString)`
- `+(SQLQueryString, SQLQueryString)`
- `[SQLQueryString].joined(separator: String)`

The test cases describe  how to use them and their usefulness.

I'm new to Vapor4 so if there already be alternative way to construct raw queries, please tell me.


t-ae:
Implemented in #112.
Thank you @gwynne !

PR #118, opened 2020-08-27
Added check to run MySQL benchmarks using TEXT field for .text

bottlehall:
There is an issue open in vapor/mysql-kit to change the underlying raw SQL type for `.text` fields to `TEXT`. However, simply making this changes causes tests to fail in SQLKitBenchmarks because not all dialects of SQL support default values for TEXT fields. @tanner0101 suggested changing the test to check for the dialect and create an appropriate column type. I think this PR does this.


bottlehall:
The issue referred to is: https://github.com/vapor/mysql-kit/pull/291

gwynne:
Closing this due to age.

PR #119, opened 2020-10-06
Case Syntax Support

calebkleveter:


0xTim:
Without being _that_ person, would be great to see some tests!

gwynne (changes requested):
Caleb, shame on you 😆

Instead of putting all these new methods on `SQLSelectBuilder`,

1) You wanted to add this to `SQLPredicateBuilder` too, so they could be used anywhere a value shows up.
2) Don't just put the whole functionality on `SELECT` columns...
3) Create an `SQLCaseBuilder` that provides methods like `func when(_ expression: SQLExpression) -> Self`, and the other `when()` methods you came up with, and the `else()` methods, and have it expose a `caseExpression` property of type `SQLExpression` that holds the built `SQLCaseExpression`.
4) In the same file, add:
```swift
extension SQLSelectBuilder {
    public func column(`case`: (SQLCaseBuilder) -> SQLCaseBuilder) -> Self {
        let builder = SQLCaseBuilder()
        _ = clauses(builder)
        return self.column(builder.caseExpression)
    }
}
```
and
```swift
extension SQLPredicateBuilder {
     public func where(`case`: (SQLCaseBuilder) -> SQLCaseBuilder) -> Self {
         // etc.
         return self.where(builder.caseExpression)
     }
}
```
And any others that seem appropriate. You can look at how `SQLPredicateGroupBuilder` and my `SQLConflictUpdateBuilder` (see the upserts PR) work for examples of how to make a new builder properly.

And like @0xTim said, add tests!

Then come back and ask for review again 😉

gwynne:
Note: In my moment of imperious tone, I forgot to offer any reason for insisting that it be done this way. My apologies - I dare say the least I could do if I'm gonna demand a rewrite is explain why 😅.

The short version is, SQLKit only remains useful as an abstraction if its facilities can be applied to handling better-structured queries, and the differences between each database's syntax, in the most general sense reasonably possible. The `CASE WHEN` syntax is an excellent addition for those who prefer using it in SQL (I'm not one of them, but that's personal preference and I still see its value), but at least in MySQL (and I believe in Postgres also), `CASE` can appear anywhere that `IF` can - in order words, as part of any expression. Adding it in a way that significantly expands the API surface of `SQLSelectBuilder`, while at the same time limiting its usage to only one specific part of the query, would not make for a maximally reusable and flexible API.

Sources/SQLKit/Builders/SQLCaseBuilder.swift
```diff
@@ -75,3 +75,16 @@ extension SQLCaseBuilder {
         return self.else(SQLBind(value))
     }
 }
+
+
+extension SQLSelectBuilder {
+    public func column(`case`: (SQLCaseBuilder) -> SQLCaseBuilder) -> Self {
+        return self.column(`case`(.init()).query)
```
gwynne, line 82: @calebkleveter Sorry, I did write it the way I originally had it for a reason - it's not technically safe to rely on the closure the user provides returning the same builder.

```suggestion
        let builder = SQLCaseBuilder()
        _ = `case`(builder)
        return self.column(builder.query)
```

Sources/SQLKit/Builders/SQLCaseBuilder.swift
```diff
@@ -75,3 +75,16 @@ extension SQLCaseBuilder {
...
+    }
+}
+
+extension SQLPredicateBuilder {
+    public func `where`(`case`: (SQLCaseBuilder) -> SQLCaseBuilder) -> Self {
+        return self.where(`case`(.init()).query)
```
gwynne, line 88:
```suggestion
        let builder = SQLCaseBuilder()
        _ = `case`(builder)
        return self.where(builder.query)
```

gwynne (changes requested):

gwynne:
Closing due to age

PR #122, opened 2021-01-09
Resurrect support for Codables with nullable fields

danramteke:
When a type conforms to `Codable`, its nullable fields will now be encoded as null, which will correctly map to nullable fields in the database table (https://github.com/vapor/sql-kit/pull/122, fixes https://github.com/vapor/sql-kit/issues/121).

PR #123, opened 2021-02-12
Revert "Resurrect support for Codables with nullable fields (#122)"

gwynne:
With apologies to @danramteke, this change reverts #122. Unfortunately, SQLKit's CI does not currently run the test suites of the various database drivers, which would have revealed that this change breaks the ability to use `GENERATED BY DEFAULT AS IDENTITY` columns (more commonly called auto-increment in other databases) in PostgreSQL.

For example, when using `SQLInsertBuilder.model(_:)`, in the past an unset `id` column in the input model would be left entirely unspecified in the resulting query, which is equivalent to specifying it explicitly with the `DEFAULT` keyword. With the `SQLQueryEncoder` changes in place, the column is given an explicit `NULL` value instead, which is not a valid input for a generated identifier column in PostgreSQL.

To make matters more confusing, `NULL` _is_ a valid input for MySQL's `AUTO_INCREMENT` and SQLite's `INTEGER PRIMARY KEY` - and of course, FluentKit does not rely on this particular piece of SQLKit for lowering models to query inputs. The result is that while this change invalidates invariants that Fluent does depend on, the problem only manifests when using PostgresKit via FluentPostgresDriver and the SQLKit layer.

Some additional study is needed to find a solution that does not invalidate Fluent's invariants. Ideally, a complete solution correctly manages the distinction between "unset non-optional", "unset optional", and "set but nil optional" values.

### DIFF
--- a/Sources/SQLKit/SQLQueryEncoder.swift
+++ b/Sources/SQLKit/SQLQueryEncoder.swift
@@ -94,29 +94,6 @@ private final class _Encoder: Encoder {
             }
         }
 
-      mutating func _encodeIfPresent<T>(_ value: T?, forKey key: Key) throws where T : Encodable {
-        if let value = value {
-          try encode(value, forKey: key)
-        } else {
-          try encodeNil(forKey: key)
-        }
-      }
-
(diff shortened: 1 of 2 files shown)


calebkleveter (approved):

tanner0101:
These changes are now available in [3.7.2](https://github.com/vapor/sql-kit/releases/tag/3.7.2)

PR #124, opened 2021-02-13
Allow Codables with nillable columns to encode as nil, if they wish

danramteke:
In #122 , we resurrected support for nillable columns, however, this caused some issues with downstream clients of this repo. 

With this release, we allow clients to opt-in to this behavior.

Here is an example:

```
struct Gas: Codable {
  let name: String
  let color: String?
}

let db = TestDatabase()
var serializer = SQLSerializer(database: db)

let insertBuilder = try db.insert(into: "gasses").model(Gas(name: "oxygen", color: nil), nilEncodingStrategy: .asNil)
insertBuilder.insert.serialize(to: &serializer)

XCTAssertEqual(serializer.sql, "INSERT INTO `gasses` (`name`, `color`) VALUES (?, NULL)")
XCTAssertEqual(serializer.binds.count, 1)
XCTAssertEqual(serializer.binds[0] as? String, "oxygen")
```

## Details

There is a new `NilEncodingStrategy` right next to the existing `KeyEncodingStrategy`. It defaults to the pre-existing behavior that works with Fluent. Clients of this library can pass in a nilEncodingStrategy option to opt-in to the nillable column behavior.

### DIFF
--- a/Sources/SQLKit/SQLQueryEncoder.swift
+++ b/Sources/SQLKit/SQLQueryEncoder.swift
@@ -94,6 +94,29 @@ private final class _Encoder: Encoder {
             }
         }
 
+      mutating func _encodeIfPresent<T>(_ value: T?, forKey key: Key) throws where T : Encodable {
+        if let value = value {
+          try encode(value, forKey: key)
+        } else {
+          try encodeNil(forKey: key)
+        }
+      }
+
+      mutating func encodeIfPresent<T>(_ value: T?, forKey key: Key) throws where T : Encodable { try _encodeIfPresent(value, forKey: key)}
+      mutating func encodeIfPresent(_ value: Int?, forKey key: Key) throws { try _encodeIfPresent(value, forKey: key) }
+      mutating func encodeIfPresent(_ value: Int8?, forKey key: Key) throws { try _encodeIfPresent(value, forKey: key) }
+      mutating func encodeIfPresent(_ value: Int16?, forKey key: Key) throws { try _encodeIfPresent(value, forKey: key) }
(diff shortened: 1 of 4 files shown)


gwynne:
This looks good to me, but I'd like an additional eye on it for good measure. @siemensikkema?

Sources/SQLKit/SQLQueryEncoder.swift
```diff
@@ -12,6 +13,13 @@ public struct SQLQueryEncoder {
         return encoder.row
     }
 
+    public enum NilEncodingStrategy {
+        /// Standard nil encoding
+        case standard
```
siemensikkema, line 18: Nit/bikeshedding: `default` seems more common in the Vapor circles and appropriate. Also, it might be good to explain here what standard (or default) _means_. Something like: "Skips nilable columns with nil values during encoding."

Sources/SQLKit/SQLQueryEncoder.swift
```diff
@@ -12,6 +13,13 @@ public struct SQLQueryEncoder {
...
+        /// Encodes nilable columns with nil values as nil. Useful when using `SQLInsertBuilder` to insert `Codable` models without Fluent
```
siemensikkema, line 19:
```suggestion
        /// Encodes nilable columns with nil values as nil. Useful when using `SQLInsertBuilder` to insert `Codable` models without Fluent.
```

siemensikkema (approved):

tanner0101:
These changes are now available in [3.8.0](https://github.com/vapor/sql-kit/releases/tag/3.8.0)

tanner0101:
These changes are now available in [3.8.0](https://github.com/vapor/sql-kit/releases/tag/3.8.0)

PR #125, opened 2021-02-22
fix select query without from

Freax13:
Not all `SELECT` queries require a `FROM` statement. This pr fixes the query generation for queries without `FROM` and allows the following example to work:

```swift
conn.sql()
    .select()
    .column(SQLAlias.init(SQLFunction("LAST_INSERT_ID"), as: SQLIdentifier.init("id")))
    .first()
```

### DIFF
--- a/Sources/SQLKit/Query/SQLSelect.swift
+++ b/Sources/SQLKit/Query/SQLSelect.swift
@@ -52,8 +52,10 @@ public struct SQLSelect: SQLExpression {
             serializer.write("DISTINCT ")
         }
         SQLList(self.columns).serialize(to: &serializer)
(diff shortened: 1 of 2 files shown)


gwynne (approved):
Huh, I coulda sworn I'd already fixed that... guess not. Thanks, looks great!

gwynne (changes requested):
@Freax13 Sorry, I just realized two seconds before I hit the merge button... Could you add a unit test for this? 😅

gwynne (approved):

tanner0101:
These changes are now available in [3.8.1](https://github.com/vapor/sql-kit/releases/tag/3.8.1)

PR #129, opened 2021-05-12
Fix raw interpolation in readme

valerianb:
Hi,

I believe the readme documentation is using the deprecated string interpolation in the raw example.
This should fix it.


0xTim (approved):
Yes it should thanks!

PR #133, opened 2021-06-17
Add support for UNION queries

finestructure:
Adds support for `UNION SELECT` queries:

```swift
try db.select()
     .column("id")
     .from("t1")
     .where("f1", .equal, "foo")
     .limit(1)
     .union({
         $0.column("id")
             .from("t2")
             .where("f2", .equal, "bar")
             .limit(2)
     }).union(all: {
         $0.column("id")
             .from("t3")
             .where("f3", .equal, "baz")
             .limit(3)
     })
     .run().wait()
```

### DIFF
--- a/Sources/SQLKit/Builders/SQLUnionBuilder.swift
+++ b/Sources/SQLKit/Builders/SQLUnionBuilder.swift
@@ -7,19 +7,15 @@ public final class SQLUnionBuilder: SQLQueryBuilder {
     public var database: SQLDatabase
 
     public init(on database: SQLDatabase,
-                _ args: [SQLSelectBuilder],
-                all: Bool = false) {
-        self.union = .init(args.map(\.select), all: all)
+                all: Bool = false,
+                _ args: [SQLSelectBuilder]) {
+        self.union = .init(all: all, args.map(\.select))
         self.database = database
     }
 }
 
 extension SQLDatabase {
-    public func union(_ args: SQLSelectBuilder...) -> SQLUnionBuilder {
-        SQLUnionBuilder(on: self, args)
-    }
-
-    public func unionAll(_ args: SQLSelectBuilder...) -> SQLUnionBuilder {
-        SQLUnionBuilder(on: self, args, all: true)
+    public func union(all: Bool = false, _ args: SQLSelectBuilder...) -> SQLUnionBuilder {
+        SQLUnionBuilder(on: self, all: all, args)
     }
 }
--- a/Sources/SQLKit/Builders/SQLUnionBuilder.swift
+++ b/Sources/SQLKit/Builders/SQLUnionBuilder.swift
@@ -1,25 +1,42 @@
 public final class SQLUnionBuilder: SQLQueryBuilder {
-    public var query: SQLExpression {
-        return self.union
-    }
+    public var query: SQLExpression { self.union }
 
     public var union: SQLUnion
     public var database: SQLDatabase
 
-    public init(on database: SQLDatabase,
-                _ args: [SQLSelectBuilder],
-                all: Bool = false) {
-        self.union = .init(args.map(\.select), all: all)
+    public init(on database: SQLDatabase, initialQuery: SQLSelect) {
+        self.union = .init(initialQuery: initialQuery)
         self.database = database
     }
+
+   public func union(distinct query: SQLSelect) -> Self {
+       self.union.add(query, all: false)
+       return self
+   }
+
+   public func union(all query: SQLSelect) -> Self {
+       self.union.add(query, all: true)
+       return self
+   }
 }
 
 extension SQLDatabase {
-    public func union(_ args: SQLSelectBuilder...) -> SQLUnionBuilder {
-        SQLUnionBuilder(on: self, args)
+    public func union(_ predicate: (SQLSelectBuilder) -> SQLSelectBuilder) -> SQLUnionBuilder {
+        return SQLUnionBuilder(on: self, initialQuery: predicate(.init(on: self)).select)
+    }
+}
+
+extension SQLUnionBuilder {
+    public func union(distinct predicate: (SQLSelectBuilder) -> SQLSelectBuilder) -> Self {
+        return self.union(distinct: predicate(.init(on: self.database)).select)
     }
 
-    public func union(all args: SQLSelectBuilder...) -> SQLUnionBuilder {
-        SQLUnionBuilder(on: self, args, all: true)
+    /// Alias the `distinct` variant so it acts as the "default".
+    public func union(_ predicate: (SQLSelectBuilder) -> SQLSelectBuilder) -> Self {
+        return self.union(distinct: predicate)
+    }
+
+    public func union(all predicate: (SQLSelectBuilder) -> SQLSelectBuilder) -> Self {
+        return self.union(all: predicate(.init(on: self.database)).select)
     }
 }
--- a/Sources/SQLKit/Builders/SQLUnionBuilder.swift
+++ b/Sources/SQLKit/Builders/SQLUnionBuilder.swift
@@ -40,3 +40,21 @@ extension SQLUnionBuilder {
         return self.union(all: predicate(.init(on: self.database)).select)
     }
 }
+
+extension SQLSelectBuilder {
+    public func union(distinct predicate: (SQLSelectBuilder) -> SQLSelectBuilder) -> SQLUnionBuilder {
+        return SQLUnionBuilder(on: self.database, initialQuery: self.select)
+            .union(distinct: predicate(.init(on: self.database)).select)
+    }
+
+    /// Alias the `distinct` variant so it acts as the "default".
+    public func union(_ predicate: (SQLSelectBuilder) -> SQLSelectBuilder) -> SQLUnionBuilder {
+        return SQLUnionBuilder(on: self.database, initialQuery: self.select)
+            .union(distinct: predicate(.init(on: self.database)).select)
+    }
+
+    public func union(all predicate: (SQLSelectBuilder) -> SQLSelectBuilder) -> SQLUnionBuilder {
+        return SQLUnionBuilder(on: self.database, initialQuery: self.select)
+            .union(all: predicate(.init(on: self.database)).select)
+    }
+}
--- a/Sources/SQLKit/Builders/SQLUnionBuilder.swift
+++ b/Sources/SQLKit/Builders/SQLUnionBuilder.swift
@@ -7,15 +7,19 @@ public final class SQLUnionBuilder: SQLQueryBuilder {
     public var database: SQLDatabase
 
     public init(on database: SQLDatabase,
-                all: Bool = false,
-                _ args: [SQLSelectBuilder]) {
-        self.union = .init(all: all, args.map(\.select))
+                _ args: [SQLSelectBuilder],
+                all: Bool = false) {
+        self.union = .init(args.map(\.select), all: all)
         self.database = database
     }
 }
 
 extension SQLDatabase {
-    public func union(all: Bool = false, _ args: SQLSelectBuilder...) -> SQLUnionBuilder {
-        SQLUnionBuilder(on: self, all: all, args)
+    public func union(_ args: SQLSelectBuilder...) -> SQLUnionBuilder {
+        SQLUnionBuilder(on: self, args)
+    }
+
+    public func union(all args: SQLSelectBuilder...) -> SQLUnionBuilder {
+        SQLUnionBuilder(on: self, args, all: true)
     }
 }
--- a/Sources/SQLKit/Builders/SQLUnionBuilder.swift
+++ b/Sources/SQLKit/Builders/SQLUnionBuilder.swift
@@ -1,0 +1,19 @@
+public final class SQLUnionBuilder: SQLQueryBuilder {
+    public var query: SQLExpression {
+        return self.union
+    }
+
+    public var union: SQLUnion
+    public var database: SQLDatabase
+
+    public init(on database: SQLDatabase, _ args: [SQLSelectBuilder]) {
+        self.union = .init(args.map { $0.select })
+        self.database = database
+    }
+}
+
+extension SQLDatabase {
+    public func union(_ args: SQLSelectBuilder...) -> SQLUnionBuilder {
+        SQLUnionBuilder(on: self, args)
+    }
+}
--- a/Sources/SQLKit/Builders/SQLUnionBuilder.swift
+++ b/Sources/SQLKit/Builders/SQLUnionBuilder.swift
@@ -9,7 +9,7 @@ public final class SQLUnionBuilder: SQLQueryBuilder {
     public init(on database: SQLDatabase,
                 _ args: [SQLSelectBuilder],
                 all: Bool = false) {
-        self.union = .init(args.map { $0.select }, all: all)
+        self.union = .init(args.map(\.select), all: all)
         self.database = database
     }
 }
--- a/Sources/SQLKit/Builders/SQLUnionBuilder.swift
+++ b/Sources/SQLKit/Builders/SQLUnionBuilder.swift
@@ -6,8 +6,10 @@ public final class SQLUnionBuilder: SQLQueryBuilder {
     public var union: SQLUnion
     public var database: SQLDatabase
 
-    public init(on database: SQLDatabase, _ args: [SQLSelectBuilder]) {
-        self.union = .init(args.map { $0.select })
+    public init(on database: SQLDatabase,
+                _ args: [SQLSelectBuilder],
+                all: Bool = false) {
+        self.union = .init(args.map { $0.select }, all: all)
         self.database = database
     }
 }
@@ -15,5 +17,9 @@ public final class SQLUnionBuilder: SQLQueryBuilder {
 extension SQLDatabase {
     public func union(_ args: SQLSelectBuilder...) -> SQLUnionBuilder {
         SQLUnionBuilder(on: self, args)
+    }
+
+    public func unionAll(_ args: SQLSelectBuilder...) -> SQLUnionBuilder {
+        SQLUnionBuilder(on: self, args, all: true)
     }
 }
--- a/Sources/SQLKit/Query/SQLUnion.swift
+++ b/Sources/SQLKit/Query/SQLUnion.swift
@@ -13,18 +13,13 @@ public struct SQLUnion: SQLExpression {
     }
 
     public func serialize(to serializer: inout SQLSerializer) {
-        self.args
-            .map { [SQLGroupExpression($0)] }
-            .joined([SQLUnionJoiner(all: self.all)])
-            .forEach { $0.serialize(to: &serializer) }
-    }
-}
-
-fileprivate struct SQLUnionJoiner: SQLExpression {
-    let all: Bool
-    
-    func serialize(to serializer: inout SQLSerializer) {
-        serializer.write("UNION\(self.all ? " ALL" : "") ")
+        let groups = self.args.map(SQLGroupExpression.init)
+        guard let first = groups.first else { return }
+        first.serialize(to: &serializer)
+        for arg in groups.dropFirst() {
+            serializer.write(all ? " UNION ALL " : " UNION ")
+            arg.serialize(to: &serializer)
+        }
     }
 }
 
--- a/Sources/SQLKit/Query/SQLUnion.swift
+++ b/Sources/SQLKit/Query/SQLUnion.swift
@@ -13,13 +13,22 @@ public struct SQLUnion: SQLExpression {
     }
 
     public func serialize(to serializer: inout SQLSerializer) {
-        let groups = self.args.map(SQLGroupExpression.init)
-        guard let first = groups.first else { return }
-        first.serialize(to: &serializer)
-        for arg in groups.dropFirst() {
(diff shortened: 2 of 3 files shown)


Tests/SQLKitTests/SQLKitTests.swift
```diff
@@ -678,4 +678,22 @@ CREATE TABLE `planets`(`id` BIGINT, `name` TEXT, `diameter` INTEGER, `galaxy_nam
...
+                .where("f1", .equal, "foo")
+                .limit(2),
+            db.select()
+                .column("id")
+                .from("t2")
+                .where("f2", .equal, "bar")
+                .limit(3)
+        ).run().wait()
+
+        XCTAssertEqual(db.results[0], "(SELECT `id` FROM `t1` WHERE `f1` = ? LIMIT 2) UNION (SELECT `id` FROM `t2` WHERE `f2` = ? LIMIT 3)")
```
finestructure, line 696: My local test with a Postgres db also checks that the binds are correctly enumerated, i.e. I can assert

```swift
        XCTAssertEqual(renderSQL(union.query),
                       #"(SELECT "id" FROM "t1" WHERE "f1" = $1 LIMIT 2) UNION (SELECT "id" FROM "t2" WHERE "f2" = $2 LIMIT 3)"#)
```

I don't think there's any way to test this with `TestDatabase` and maybe that's not an error that could occur. Just wanted to point it out!

Sources/SQLKit/Query/SQLUnion.swift
```diff
@@ -0,0 +1,22 @@
...
+    public init(_ args: [SQLExpression]) {
+        self.args = args
+    }
+
+    public func serialize(to serializer: inout SQLSerializer) {
+        let groups = args.map(SQLGroupExpression.init)
+        guard let first = groups.first else { return }
+        first.serialize(to: &serializer)
+        for arg in groups.dropFirst() {
+            serializer.write(" UNION ")
```
finestructure, line 17: Just realised there's also `UNION ALL`, at least on Postgres. Not sure if it's universal across supported SQL flavours but I guess it couldn't hurt to support this via a flag defaulting to off.

Sources/SQLKit/Builders/SQLUnionBuilder.swift
```diff
@@ -0,0 +1,25 @@
...
+}
+
+extension SQLDatabase {
+    public func union(_ args: SQLSelectBuilder...) -> SQLUnionBuilder {
+        SQLUnionBuilder(on: self, args)
+    }
+
+    public func unionAll(_ args: SQLSelectBuilder...) -> SQLUnionBuilder {
+        SQLUnionBuilder(on: self, args, all: true)
+    }
```
finestructure, line 24: I'm not sure if `unionAll` is great but my thinking was that for the main API it'd be better if the `all` was more obvious than a trailing parameter.

Of course, it could instead be `union(all: true, _ args: ...)` but it feels like a bit of a weird interface in general with a boolean flag that isn't very descriptive.

Maybe it should be `union(allowDuplicates: true, _ args ...)`? Although I don't think there's API precedent to interpret SQL semantics rather than just pass through the raw syntax in a swifty way.

finestructure: I think I've convinced myself to change this to `union(all: true, _ args: ...)` (and move `all:` to the front in `SQLUnion` accordingly).

What do you think?

gwynne: How about `union(all: ...)`?

Sources/SQLKit/Builders/SQLUnionBuilder.swift
```diff
@@ -0,0 +1,21 @@
...
+    public init(on database: SQLDatabase,
+                all: Bool = false,
+                _ args: [SQLSelectBuilder]) {
+        self.union = .init(all: all, args.map(\.select))
+        self.database = database
+    }
+}
+
+extension SQLDatabase {
+    public func union(all: Bool = false, _ args: SQLSelectBuilder...) -> SQLUnionBuilder {
```
gwynne, line 18: That wasn't what I meant 😅 I was thinking more this:

```swift
public func union(_ args: SQLSelectBuilder...) -> SQLUnionBuilder
public func union(all args: SQLSelectBuilder...) -> SQLUnionBuilder
```
The different labels are sufficient to differentiate the overloads.

finestructure: Ah gotcha! If you think that's clear enough as a distinction - I like it!

gwynne: At the very least, it's a pattern Tanner and I have used repeatedly elsewhere 🤷‍♀️

0xTim: Who's Tanner? 😈

Sources/SQLKit/Query/SQLUnion.swift
```diff
@@ -0,0 +1,24 @@
...
+
+    public func serialize(to serializer: inout SQLSerializer) {
+        let groups = args.map(SQLGroupExpression.init)
+        guard let first = groups.first else { return }
+        first.serialize(to: &serializer)
+        for arg in groups.dropFirst() {
+            serializer.write(all ? " UNION ALL " : " UNION ")
+            arg.serialize(to: &serializer)
+        }
+    }
```
gwynne, line 22: I recommend doing something like this for arg handling and serialization so that we 1) don't silently accept a nonsensical construction like `query.union()`, 2) we can more easily support additional syntax for databases that may have it (like PostgreSQL's `UNION | INTERSECT | EXCEPT [ALL | DISTINCT]` syntax), and 3) maintain general composability. (Oh, and 4) consistent use of `self.` for accessing properties.)

```suggestion
    public init(_ args: [SQLExpression], all: Bool = false) {
        precondition(!args.isEmpty, "Empty SQLUnions are not valid.") 
        self.args = args
        self.all = all
    }

    public func serialize(to serializer: inout SQLSerializer) {
        self.args
            .map { [SQLGroupExpression($0)] }
            .joined([SQLUnionJoiner(all: self.all)])
            .forEach { $0.serialize(to: &serializer) }
    }
}

fileprivate struct SQLUnionJoiner: SQLExpression {
    let all: Bool
    
    func serialize(to serializer: inout SQLSerializer) {
        serializer.write("UNION\(self.all ? " ALL" : "") ")
    }
```

Sources/SQLKit/Builders/SQLUnionBuilder.swift
```diff
@@ -0,0 +1,25 @@
...
+    public init(on database: SQLDatabase,
+                _ args: [SQLSelectBuilder],
+                all: Bool = false) {
+        self.union = .init(args.map(\.select), all: all)
+        self.database = database
+    }
+}
+
+extension SQLDatabase {
+    public func union(_ args: SQLSelectBuilder...) -> SQLUnionBuilder {
```
gwynne, line 18: Elsewhere that we deal with subqueries, we tend to use this kind of construct instead of accepting builders as parameters:
```swift
extension SQLSelectBuilder {
    public func union(select builder: (SQLSelectBuilder) -> SQLSelectBuilder) -> SQLUnionBuilder {
        return .init(on: self, [self.select, predicate(.init(on: self.database))])
    }
}
```
With usage then looking like this:
```swift
database.select()
    .columns(whatever)
    .from(whatever)
.union(select: { $0
    .columns(whatever)
    .from(whatever)
})
// And an extension on SQLUnionBuilder would allow chaining this way instead of nesting closures:
.unionAll(select: { $0
    .columns(whatever)
    .from(whatever)
})
```
It could also be set up to be used like this:
```swift
database.union(select: {
    $0.columns().etcetera()
}).union(select: {
    $0.etcetera()
}).unionAll(select: {
    $0.etcetera()
})
```
Or both, which is probably what I'd go with if I were implementing this syntax.

Whether this is better or worse than the way it is here of just taking a variadic array of select builders is of course a matter of opinion; I'm open to arguments in both directions.

gwynne (changes requested):
Apologies for the second round of changes requested; I had a chance to take a closer look than I did the first time. I'm only actually "requesting changes" in the serialization logic; my thoughts about alternative usage syntax are more of just a "what does everyone think?" thing, I won't block merging just for that 😅 (Though we should have a solid decision before merging, since changing it later would be a public API break.)

Sources/SQLKit/Builders/SQLUnionBuilder.swift
finestructure: Interesting, it hadn't even occurred to me that this would be an option. I think I'd still lean towards `union(_ args: SQLSelectBuilder...`, because the operands are "equal".

What I mean is that while `.union(select: ...)` more closely matches the order of the SQL syntax, it does introduce some nesting (for lack of a better term) that the SQL syntax doesn't have. Or said differently, to me `a union b` maps more naturally to `union(a, b)` than `a.union(b)`.

This is just a personal preference though, happy to adopt whatever is preferred!

Sources/SQLKit/Query/SQLUnion.swift
finestructure: Sorry, I had to revert the `map/joined/forEach` part because it didn't actually compile and it wasn't immediately obvious how to fix it. I've kept the precondition and the `self.`.

gwynne: Sorry about that! That's what happens when I start writing code in GitHub comments without actually trying it 😅. I'll fix the errors and repost shortly.

finestructure: No worries, Compiler As A Service 🤣

Sources/SQLKit/Builders/SQLUnionBuilder.swift
gwynne: Yeah, at the very least I'd definitely lean towards the second example, where `union(select:)` etc. are methods on both `SQLDatabase` and `SQLUnionBuilder`, but not `SQLSelectBuilder`, and I also agree that your original proposal is more straightforward - my main argument for the closure-based syntax is that it matches other usages in `SQLKit`, and I tend to want to avoid making this layer any more finicky to use than it already is 😅. Of course, at the same time, "straightforward" may be better than "consistent", so all in all I'm not sure what's best either.

Sources/SQLKit/Query/SQLUnion.swift
```diff
@@ -0,0 +1,25 @@
...
+
+    public func serialize(to serializer: inout SQLSerializer) {
+        let groups = self.args.map(SQLGroupExpression.init)
+        guard let first = groups.first else { return }
+        first.serialize(to: &serializer)
+        for arg in groups.dropFirst() {
+            serializer.write(all ? " UNION ALL " : " UNION ")
+            arg.serialize(to: &serializer)
+        }
```
gwynne, line 22: Well! Turns out I made no less than _three_ mistakes originally:

1. Forgot the `separator:` label in `joined(separator:)`.
2. Forgot the compiler would not be able to infer the common `SQLExpression` type for the `forEach(_:)` closure.
3. Didn't put a space before `UNION` in the joiner 🤣.

Here's a version that both compiles and passes tests (I also made the helper struct public so I wouldn't be inconsistent with my remarks about consistency 😂):

```suggestion
        self.args
            .map { [SQLGroupExpression($0)] }
            .joined(separator: [SQLUnionJoiner(all: self.all)])
            .forEach { (item: SQLExpression) in item.serialize(to: &serializer) }
    }
}

public struct SQLUnionJoiner: SQLExpression {
    public var all: Bool
    
    public init(all: Bool) {
        self.all = all
    }
    
    public func serialize(to serializer: inout SQLSerializer) {
        serializer.write(" UNION\(self.all ? " ALL" : "") ")
```

Sources/SQLKit/Builders/SQLUnionBuilder.swift
finestructure: So if I understand correctly, this would transform the test from

```swift
        try db.union(
            db.select()
                .column("id")
                .from("t1")
                .where("f1", .equal, "foo")
                .limit(1),
            db.select()
                .column("id")
                .from("t2")
                .where("f2", .equal, "bar")
                .limit(2),
            db.select()
                .column("id")
                .from("t3")
                .where("f3", .equal, "baz")
                .limit(3)
        ).run().wait()
```

to

```swift
        try db
	.union(select: {
            $0.column("id")
                .from("t1")
                .where("f1", .equal, "foo")
                .limit(1)
		})
	.union(select: {
            $0.column("id")
                .from("t2")
                .where("f2", .equal, "bar")
                .limit(2)
		})
	.union(select: {
            $0.column("id")
                .from("t3")
                .where("f3", .equal, "baz")
                .limit(3)
        }).run().wait()
```

We're using SQLKit directly (instead of Fluent) in a few places and I've not yet come across this kind of closure based API, I think. As an API user I'd definitely prefer the former (unsurprisingly, as it's the API I reached for coming from the other end!).

In particular, I'm testing the individual union queries (which can be quite complicated) and it's helpful that they're valid standalone queries (like the original SQL queries are just concatenations). For example:

- https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/blob/swift-5.5/Sources/App/Core/Search.swift#L248-L249
- https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/blob/swift-5.5/Sources/App/Core/Search.swift#L198
- https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/blob/swift-5.5/Tests/AppTests/SearchTests.swift#L70

I'm not sure that kind of composability could be preserved unless `union` operates on builders? I have to admit I haven't spent a lot of time investigating. I'm happy to adopt and deal with whatever the API will end up being, even if it means to add a little wrapper in SPI to make that feasible.

gwynne: I admit there's only one place where closures are used by `SQLKit` already - `SQLPredicateBuilder.where(group:)` and `orWhere(group:)`, e.g. `db.select()./* blah */.where(\Baz.baz != nil).where { $0.where(\Foo.foo == 1).orWhere(\Bar.bar == 2) }` -> `WHERE baz IS NOT NULL AND (Foo.foo=1 OR Bar.bar=2)`. I also used them in my proposed syntax for providing "`UPSERT`" support (the PR is stale but not abandoned: #90).

The closure-based syntax does ultimately still operate on `SQLSelectBuilder`s; the main benefit I see from the point of view of the API caller is that they're no longer manually invoking `db.select()` inside another `db.<something>()` call - one particular benefit of having the select builders created internally by the union builder is a user can no longer (accidentally or otherwise) do something like `db.union(someOtherDb.select().whatever())` (which doesn't make sense and wouldn't have whatever effect may have been intended - even done accidentally it'd just be confusion waiting to happen for code maintainers).

0xTim: So I have no clear winner in my mind. I agree that option a above looks nicer and seems like it would compose better but I'm definitely in favour of trying to enforce some type-safety around the DB being used etc. 

So in my mind - we see if it's possible to extract out the select builder into something that can be composed (outside of just pulling the closure out into a function - is this an option?) and if not pick option 1

gwynne: The select builder can be composed directly as long as it's the closure return value as well as its parameter so the closure can return a composite of its own. The call site, in turn, should be taking the `.select` of the return value rather than of the parameter, rather than assuming they're always the same as `SQLPredicateGroupBuilder` currently incorrectly does (this admittedly partially defeats the "prevent the user from mixing databases in the builders" benefit, but composite sub-builders are a very rare usage and one would have to go out of their way to make the error rather than it being relatively trivial).

gwynne: Let me show you really quickly a sketch of how I'd implement this...

```swift
public final class SQLUnionBuilder: SQLQueryBuilder {
     public var query: SQLExpression { self.union }

     public var union: SQLUnion
     public var database: SQLDatabase

     public init(on database: SQLDatabase, initialQuery: SQLSelect) {
         self.union = .init(initialQuery: initialQuery)
         self.database = database
     }
    
    public func union(distinct query: SQLSelect) -> SQLUnionBuilder {
        self.union.add(query, all: false)
        return self
    }
    
    public func union(all query: SQLSelect) -> SQLUnionBuilder {
        self.union.add(query, all: true)
        return self
    }
 }

public struct SQLUnion: SQLExpression {
    public var initialQuery: SQLSelect
    public var unions: [(SQLUnionJoiner, SQLSelect)]
    
    public init(initialQuery: SQLSelect, unions: [(SQLUnionJoiner, SQLSelect)] = []) {
        self.initialQuery = initialQuery
        self.unions = unions
    }
    
    public func add(_ query: SQLSelect, all: Bool) {
        self.unions.append((.init(all: all), query))
    }

    public func serialize(into serializer: inout SQLSerializer) {
        assert(!self.unions.isEmpty, "Serializing a union with only one query is invalid.")
        self.initialQuery.serialize(into: &serializer)
        self.unions.flatMap { [$0, $1] }.forEach( (item: SQLExpression) in item.serialize(into: &serializer) }
    }
}

extension SQLDatabase {
    public func union(_ predicate: (SQLSelectBuilder) -> SQLSelectBuilder) -> SQLUnionBuilder {
        return SQLUnionBuilder(on: self, initialQuery: predicate(.init(on: self)).select)
    }
}

extension SQLUnionBuilder {
    public func union(distinct predicate: (SQLSelectBuilder) -> SQLSelectBuilder) -> Self {
        return self.union(distinct: predicate(.init(on: self.database)).select)
    }

    public func union(all predicate: (SQLSelectBuilder) -> SQLSelectBuilder) -> Self {
        return self.union(all: predicate(.init(on: self.database)).select)
    }
}
```

Something along those lines, anyhow.

finestructure: This snippet:

```swift
    public func serialize(into serializer: inout SQLSerializer) {
        assert(!self.unions.isEmpty, "Serializing a union with only one query is invalid.")
        self.initialQuery.serialize(into: &serializer)
        self.unions.flatMap { [$0, $1] }.forEach( (item: SQLExpression) in item.serialize(into: &serializer) }
    }
```

didn't quite compile. I've come up with this replacement that passes the tests (will push a change in a minute):

```swift
        assert(!self.unions.isEmpty, "Serializing a union with only one query is invalid.")
        SQLGroupExpression(self.initialQuery).serialize(to: &serializer)
        self.unions
            .forEach { (joiner, select) in
                joiner.serialize(to: &serializer)
                SQLGroupExpression(select).serialize(to: &serializer)
            }
    }
```

Was that what you had in mind? It looks quite different, so I'm not sure.

Tests/SQLKitTests/SQLKitTests.swift
```diff
@@ -717,4 +717,41 @@ CREATE TABLE `planets`(`id` BIGINT, `name` TEXT, `diameter` INTEGER, `galaxy_nam
...
+            $0.column("id")
+                .from("t2")
+                .where("f2", .equal, "bar")
+                .limit(2)
+        }).union({
+            $0.column("id")
+                .from("t3")
+                .where("f3", .equal, "baz")
+                .limit(3)
+        })
```
finestructure, line 737: Note that

```swift
db.union {
}
.union {
}
...
```

fails with `Ambiguous use of 'union'` in the first line. Subsequent closures need to be placed in `(...)`.

finestructure: This can be somewhat mitigated by adding

```swift
extension SQLSelectBuilder {
    public func union(distinct predicate: (SQLSelectBuilder) -> SQLSelectBuilder) -> SQLUnionBuilder {
        return SQLUnionBuilder(on: self.database, initialQuery: self.select)
            .union(distinct: predicate(.init(on: self.database)).select)
    }

    /// Alias the `distinct` variant so it acts as the "default".
    public func union(_ predicate: (SQLSelectBuilder) -> SQLSelectBuilder) -> SQLUnionBuilder {
        return SQLUnionBuilder(on: self.database, initialQuery: self.select)
            .union(distinct: predicate(.init(on: self.database)).select)
    }

    public func union(all predicate: (SQLSelectBuilder) -> SQLSelectBuilder) -> SQLUnionBuilder {
        return SQLUnionBuilder(on: self.database, initialQuery: self.select)
            .union(all: predicate(.init(on: self.database)).select)
    }
}
```

allowing:

```swift
db.select()
            .column("id")
            .from("t1")
            .where("f1", .equal, "foo")
            .limit(1)
            .union({
                $0.column("id")
                    .from("t2")
                    .where("f2", .equal, "bar")
                    .limit(2)
            }).union(all: {
                $0.column("id")
                    .from("t3")
                    .where("f3", .equal, "baz")
                    .limit(3)
            })
```

I'll commit that as well now.

Overall, I think the closure based syntax is a little clunkier in use but on the other hand it allows expressing `... UNION ... UNION ALL ...`. I don't think that's possible with my initially proposed API. (You can probably nest queries to work around that though.)

The biggest problem for me is the implementation. I understand maybe 64% of it today, and that understanding will decay rapidly for here. I hope it's going to make sense for future maintainers 😅

finestructure: @gwynne , @0xTim - just checking - are there any further changes required to get this merged? :)

gwynne (approved):
Nope, LGTM at this point!

PR #140, opened 2021-11-20
Add Concurrency extensions for SQLKit

gwynne:
Shockingly little effort required here! SQLKit doesn't do very much asynchronous work, in and of itself.


jdmcd (approved):
LGTM, unfortunate that the tests won't run on Linux but ¯\_(ツ)_/¯

PR #142, opened 2021-11-27
Fix Swift compiler choosing wrong overload for SQLInsertBuilder.values(Encodable)

gwynne:
Under certain specific circumstances, the Swift compiler will choose the wrong overload of `SQLInsertBuilder.values(_:)` for the variants that take `Encodable` (specifically when an explicit Array is used in a generic context). Since it's too late to fix the API (it's public now), adding @_disfavoredOverload works around the issue in what should be a backwards-compatible fashion.

### DIFF
--- a/Tests/SQLKitTests/SQLKitTests.swift
+++ b/Tests/SQLKitTests/SQLKitTests.swift
@@ -304,6 +304,19 @@ final class SQLKitTests: XCTestCase {
         
         XCTAssertEqual(db.results[0], "UPDATE `planets` SET `moons` = `moons` + 1 WHERE `best_at_space` >= ?")
     }
+    
+    func testInsertWithArrayOfEncodable() throws {
+        func weird<S: Sequence>(_ builder: SQLInsertBuilder, values: S) -> SQLInsertBuilder where S.Element: Encodable {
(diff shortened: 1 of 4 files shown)


Tests/SQLKitTests/SQLKitTests.swift
```diff
@@ -304,6 +304,19 @@ final class SQLKitTests: XCTestCase {
...
     }
+    
+    func testInsertWithArrayOfEncodable() throws {
+        func weird<S: Sequence>(_ builder: SQLInsertBuilder, values: S) -> SQLInsertBuilder where S.Element: Encodable {
+            builder.values(Array(values))
+        }
+        
+        try weird(db.insert(into: "planets")
+            .columns("name")
+            , values: ["Jupiter"])
```
jdmcd, line 315: Extra comma? Or at least weird spacing

jdmcd (changes requested):

Tests/SQLKitTests/SQLKitTests.swift
gwynne: Weird spacing 😅 Fixed

jdmcd (approved):

PR #145, opened 2022-02-04
Adds `ALTER _ RENAME TO _` support

NeedleInAJayStack:
Adds table-renaming support to `SQLAlterTableBuilder`. This produces SQL queries of the form:

```sql
ALTER `table_name` RENAME TO `new_table_name`
```

### DIFF
--- a/Sources/SQLKit/Builders/SQLAlterTableBuilder.swift
+++ b/Sources/SQLKit/Builders/SQLAlterTableBuilder.swift
@@ -20,6 +20,21 @@ public final class SQLAlterTableBuilder: SQLQueryBuilder {
     }
     
     @discardableResult
+    public func rename(
+        to newName: String
+    ) -> Self {
+        return self.rename(to: SQLIdentifier(newName))
+    }
+    
+    @discardableResult
+    public func rename(
+        to newName: SQLExpression
+    ) -> Self {
+        self.alterTable.renameTo = newName
+        return self
+    }
+    
+    @discardableResult
     public func column(
         _ column: String,
         type dataType: SQLDataType,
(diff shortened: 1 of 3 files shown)


Sources/SQLKit/Builders/SQLAlterTableBuilder.swift
```diff
@@ -19,6 +19,21 @@ public final class SQLAlterTableBuilder: SQLQueryBuilder {
         self.database = database
     }
     
+    @discardableResult
```
0xTim, line 22: Could you add some API docs for this? I know the rest of the class doesn't have this but we should make sure we try and add them as we go forward

Tests/SQLKitTests/SQLKitTests.swift
```diff
@@ -209,47 +209,52 @@ final class SQLKitTests: XCTestCase {
 
     func testAltering() throws {
         // SINGLE
+        try db.alter(table: "alterable")
+            .rename(to: "new_alterable")
+            .run().wait()
+        XCTAssertEqual(db.results[0], "ALTER TABLE `alterable` RENAME TO `new_alterable`")
+
```
0xTim, line 216: Could you move this to the bottom of the function, would revert changing the existing asserts so it's clear they don't change them

0xTim (changes requested):
Thanks for this! Looks good, just a couple of small changes requested

Sources/SQLKit/Builders/SQLAlterTableBuilder.swift
NeedleInAJayStack: Sure, no problem. I've added API docs throughout this file for good measure.

Tests/SQLKitTests/SQLKitTests.swift
NeedleInAJayStack: Definitely. Done.

0xTim (approved):
LGTM - @gwynne any last thoughts?

gwynne (approved):
LGTM too!

PR #147, opened 2022-04-08
Adds `LIMIT`, `OFFSET` and `ORDER BY` to Union results

NeedleInAJayStack:
Previously it was not easy to do pagination on the result of `SELECT ... UNION queries`. For example:

```sql
(SELECT * FROM "Table" WHERE "name" = 'first thing')
UNION ALL
(SELECT * FROM "Zone" WHERE "name" = 'second thing')
LIMIT 5
OFFSET 3
ORDER BY "name"
``` 

This pull request adds `LIMIT`, `OFFSET`, and `ORDER BY` functionality to the `SQLUnion`/`SQLUnionBuilder`, primarily by copying over the `SQLSelect` and `SQLSubqueryClauseBuilder` implementations.

I also have a protocol-based approach that reduces code copying but it comes with it's own smells. I'm happy to elaborate if desired.

### DIFF
--- a/Sources/SQLKit/Query/SQLUnion.swift
+++ b/Sources/SQLKit/Query/SQLUnion.swift
@@ -1,10 +1,22 @@
 public struct SQLUnion: SQLExpression {
     public var initialQuery: SQLSelect
     public var unions: [(SQLUnionJoiner, SQLSelect)]
+    
+    /// Zero or more `ORDER BY` clauses.
+    public var orderBys: [SQLExpression]
+    
+    /// If set, limits the maximum number of results.
+    public var limit: Int?
+    
+    /// If set, offsets the results.
+    public var offset: Int?
 
     public init(initialQuery: SQLSelect, unions: [(SQLUnionJoiner, SQLSelect)] = []) {
         self.initialQuery = initialQuery
         self.unions = unions
+        self.limit = nil
+        self.offset = nil
+        self.orderBys = []
     }
 
     public mutating func add(_ query: SQLSelect, all: Bool) {
@@ -33,6 +45,19 @@ public struct SQLUnion: SQLExpression {
             self.unions.forEach { joiner, query in
                 statement.append(joiner)
                 appendQuery(query)
+            }
+            
+            if !self.orderBys.isEmpty {
+                statement.append("ORDER BY")
+                statement.append(SQLList(self.orderBys))
+            }
+            if let limit = self.limit {
+                statement.append("LIMIT")
+                statement.append(limit.description)
+            }
+            if let offset = self.offset {
+                statement.append("OFFSET")
+                statement.append(offset.description)
             }
         }
     }
--- a/Tests/SQLKitTests/SQLKitTests.swift
+++ b/Tests/SQLKitTests/SQLKitTests.swift
@@ -73,7 +73,19 @@ final class SQLKitTests: XCTestCase {
             .wait()
         XCTAssertEqual(db.results[0], "SELECT LAST_INSERT_ID() AS `id`")
     }
-
+    
+    func testSelect_limitAndOrder() throws {
+        try db.select()
+            .column("*")
+            .from("planets")
+            .limit(3)
+            .offset(5)
+            .orderBy("name")
+            .run()
+            .wait()
+        XCTAssertEqual(db.results[0], "SELECT * FROM `planets` ORDER BY `name` ASC LIMIT 3 OFFSET 5")
+    }
+    
     func testUpdate() throws {
         try db.update("planets")
             .where("name", .equal, "Jpuiter")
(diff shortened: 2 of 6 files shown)


gwynne (changes requested):
I like this change, but whenever I find that the same methods are being repeatedly written on multiple different builders in SQLKit, I tend to look for whether those methods would be better placed on one of the common builder protocols (such as `SQLPredicateBuilder`), or in some cases whether a new builder protocol is warranted. In this case, I'd go with a new `SQLPartialResultBuilder` protocol which includes the limit, offset, and order by clauses that both `SQLSelectBuilder` and `SQLUnionBuilder` conform to.

NeedleInAJayStack:
Hey @gwynne, thanks for the review!

I've broken out a builder protocol that `SQLSelectBuilder` and `SQLUnionBuilder` both conform to. I thought that `SQLPaginateBuilder` was more descriptive than `SQLPartialResultBuilder`, but I'm definitely happy to change it to `SQLPartialResultBuilder` if you prefer.

Thanks again!

gwynne:
> I've broken out a builder protocol that `SQLSelectBuilder` and `SQLUnionBuilder` both conform to. I thought that `SQLPaginateBuilder` was more descriptive than `SQLPartialResultBuilder`, but I'm definitely happy to change it to `SQLPartialResultBuilder` if you prefer.

I would prefer the more generic name in this case, mostly because it _is_ more generic and the intent of the various builder protocols is to encapsulate generalized functionality which is common to multiple - but not all - builder contexts; pagination is a very specific operation that is supported by the clauses in question, but not the only one (to grab a trivial example, the `LIMIT` clause is also used to ensure there's only one result from a `.first()` operation).

Other than that, great work! I added a couple of minor additional remarks to the updated code, once that's all taken care of, this should be ready for merging 👍

Sources/SQLKit/Builders/SQLPaginateBuilder.swift
```diff
@@ -0,0 +1,72 @@
+
+public protocol SQLPaginatableBuilder: AnyObject {
+    /// Zero or more `ORDER BY` clauses.
+    var orderBy: [SQLExpression] { get set }
```
gwynne, line 4: Can we call this property `orderBys`? It's more grammatically awkward for sure, but it avoids any naming confusion versus the `orderBy()` methods and matches the convention used on other builder protocols like `SQLJoinBuilder`

gwynne:
I wish the API breakage checker was smarter about cross-protocol refactoring 😞

Sources/SQLKit/Builders/SQLPaginateBuilder.swift
NeedleInAJayStack: Sure. I've changed the `SQLPartialResultBuilder` var (and therefore the conforming `SQLPartialResultBuilder` and `SQLUnionBuilder` vars) to `orderBys`. I also changed the `SQLUnion` var to `orderBys` because it matches the convention of `unions`. The `SQLSelect` `orderBy` var is already part of the public API though so I left that one as is.

Sources/SQLKit/Query/SQLUnion.swift
```diff
@@ -35,6 +47,18 @@ public struct SQLUnion: SQLExpression {
...
+            SQLList(self.orderBys).serialize(to: &serializer)
+        }
+        if let limit = self.limit {
+            serializer.write(" LIMIT ")
+            serializer.write(limit.description)
+        }
+        if let offset = self.offset {
+            serializer.write(" OFFSET ")
+            serializer.write(offset.description)
+        }
```
gwynne, line 61: Whoops, I missed this part before- Can you move these writes into the `serializer.statement {` closure and update them as per below? GitHub doesn't let me apply a code suggestion to lines that weren't changed in the diff, so I can't quite show it cleanly. (Also, someone should really fix `SQLSelect`'s serialization to at least properly use a numeric literal expression, but that's out of scope for this PR.)
```suggestion
            if !self.orderBys.isEmpty {
                statement.append("ORDER BY")
                statement.append(SQLList(self.orderBys))
            }
            if let limit = self.limit {
                statement.append("LIMIT")
                statement.append(SQLLiteral.numeric("\(limit)"))
            }
            if let offset = self.offset {
                statement.append("OFFSET")
                statement.append(SQLLiteral.numeric("\(offset)"))
            }
```

Tests/SQLKitTests/SQLKitTests.swift
```diff
@@ -917,5 +929,19 @@ CREATE TABLE `planets`(`id` BIGINT, `name` TEXT, `diameter` INTEGER, `galaxy_nam
...
+        // Test that LIMIT, OFFSET, and ORDERBY are applied correctly
+        db._dialect.unionFeatures.remove(.explicitDistinct)
+        try db.select()
+            .column("id").from("t1")
+            .union({
+                $0.column("id").from("t2")
+            })
+            .limit(3)
+            .offset(5)
+            .orderBy("id")
```
gwynne, line 942: While I'm being pedantic, can you add a test that makes sure multiple `ORDER BY` columns are serialized correctly as well?

gwynne:
(Sorry that I keep finding more things to fix; I haven't looked closely enough every time through, I'm afraid 🤦‍♀️)

Tests/SQLKitTests/SQLKitTests.swift
NeedleInAJayStack: Sure, I've added one.

Sources/SQLKit/Query/SQLUnion.swift
NeedleInAJayStack: Oh totally, no problem. Yeah that makes sense with SQLSelect, since I definitely just based this code off that one.

NeedleInAJayStack:
> (Sorry that I keep finding more things to fix; I haven't looked closely enough every time through, I'm afraid 🤦‍♀️)

No worries at all! It's better to get it right now than have to fix it later.

gwynne (approved):
Looking great! We can ignore the failures in macOS CI, it's just Concurrency in XCTest causing false TSan positives again; the failures disappear in Xcode 14 (though whether TSan got bugfixes or it's due to SE-0338 taking effect, I'm not sure 😅).

PR #148, opened 2022-04-11
UNION handles single entry

NeedleInAJayStack:
This PR simplifies dynamically building UNION statements.

To be specific, it loosens the restriction that a SQLUnion must contain multiple select statements. While this is common usage, it makes building up `UNION`s in client code difficult. For example, building up a UNION in a for loop is awkward right now:

```swift
let ids = [1, 2, 3, ...]
guard let firstId = ids.first else { ... }

// Must manually short-circuit as a SQLSelectBuilder
guard ids.count > 1 else {
    return sql.select.column("id").from("t1").where("id", .equals, firstId).all()
}
let unionBuilder = sql.union { select in
    select.column("id").from("t1").where("id", .equals, firstId)
}
for id in ids[1..<ids.count] {
    unionBuilder.union(all: { select in
        select.column("id").from("t1").where("id", .equals, id)
    })
}
return unionBuilder.all()
```

This PR removes the need for the commented `guard` in the code above. It also improves code safety by removing a runtime fatal error condition.

### DIFF
--- a/Sources/SQLKit/Query/SQLUnion.swift
+++ b/Sources/SQLKit/Query/SQLUnion.swift
@@ -16,7 +16,9 @@ public struct SQLUnion: SQLExpression {
     }
 
     public func serialize(to serializer: inout SQLSerializer) {
-        assert(!self.unions.isEmpty, "Serializing a union with only one query is invalid.")
+        guard !self.unions.isEmpty else {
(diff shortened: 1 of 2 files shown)


0xTim (approved):
LGTM, but @gwynne is the person who knows and owns this code!

gwynne (approved):
LGTM!

PR #152, opened 2022-07-20
Add ability for SQLKit drivers to expose version numbers

gwynne:
Also fills in some missing documentation.

### DIFF
--- a/Sources/SQLKit/SQLDatabase.swift
+++ b/Sources/SQLKit/SQLDatabase.swift
@@ -1,7 +1,49 @@
+import Logging
+import NIOCore
+
+/// The core of an SQLKit driver. This common interface is the access point of both SQLKit itself and
+/// SQLKit clients to all of the information and behaviors necessary to provide and leverage the
+/// package's functionality.
+///
+/// Conformances to ``SQLDatabase`` are typically provided by an external database-specific driver
+/// package, alongside a few utility wrapper types for handling deferred and pooled connection
+/// logic and for substituting ``Logger``s. A driver package must also provide concrete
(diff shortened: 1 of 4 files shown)


Sources/SQLKit/SQLDatabase.swift
```diff
@@ -1,14 +1,94 @@
...
+    /// may also change at runtime (for example, if a connection is auto-reconnected after a remote
+    /// update), or even become unknown again after being known.
+    ///
+    /// - Warning: This version number has nothing to do with ``SQLKit`` or (usually) of the driver
+    ///   implementation for the database, nor does it represent any data stored within the database;
+    ///   it is the version of the database implementation _itself_ (such as of a MySQL server or
+    ///   `libsqlite3` library). A significant part of the motivation to finally add this property comes
+    ///   from a larger desire to enable customizing a given ``SQLDialect``'s configuration based on the
+    ///   actual feature set available at runtime instead of having to hardcode a "safe" baseline.
+    var version: SQLDatabaseReportedVersion? { get }
```
0xTim, line 59: Is it worth renaming this to `databaseVersion` or similar to avoid any ambiguity

0xTim (approved):
LGTM - one query around naming but take it or leave it

Sources/SQLKit/SQLDatabase.swift
gwynne: I'm already unhappy about the repetition of "database" in the name of the type, honestly 😕

0xTim: Fair - I'm just thinking of potential future versions (no pun intended) where we expose other versions, but I'm happy with this as is

PR #160, opened 2022-12-18
Add configuration option for dialects without multi row insert support

lovetodream:
This change adds a configuration parameter (`SQLDialect.supportsMultiRowInsert`) to support dialects that do not support multi row inserts (e.g. Oracle). This is `true` by default. If `false`, a warning will be logged once multiple rows are about to be inserted into the database within a single statement.


lovetodream:
It might be more appropriate to throw an error, because the statement will most likely fail anyways. I'd like to hear some opinions about this.

gwynne:
Closing this out due to age and the existing difficulty of dealing with lack of support for various other features by specific databases.

PR #161, opened 2023-01-11
Allow to pass keyDecodingStrategy in first(decoding:) and all(decoding:)

valerianb:
When inserting, you can specify a keyEncodingStrategy:
```
try db.insert(into: myTable)
  .model(myModel, keyEncodingStrategy: .convertToSnakeCase)
  .run()
```
This change allows to pass a keyDecodingStrategy when using the convenience methods to decode models directly:
```
db.select()
  .from(myTable)
  .all(decoding: MyModel.self, keyDecodingStrategy: .convertFromSnakeCase)
```


gwynne:
This was addressed by #172

PR #166, opened 2023-05-05
Major codebase cleanup

gwynne:
A massive pass across the entirety of SQLKit, doing the following:

- Fix, correct, clarify, add, and streamline documentation comments across the board.
- Use `any` in all places it should appear.
- Normalize coding style (esp. omitting unneeded `return`s and consistent use of `self.`).
- Add `@inlinable` and `@usableFromInline` to a significant amount of the API, hopefully improving performance.
- Fix some minor copy-pasta bugs, mostly in `SQLCreateTrigger`.
- Deprecate use of non-trivial raw SQL strings outside of `SQLQueryString` or `SQLRaw`.


README.md
```diff
@@ -1,55 +1,54 @@
...
 
+SQLKit 3.x requires [SwiftNIO](https://github.com/apple/swift-nio) 2.x or later. Previous major versions are no longer supported.
+
 ### Supported Platforms
 
-PostgresNIO supports the following platforms:
+SQLKit supports the following platforms:
 
-- Ubuntu 16.04+
+- Ubuntu 18.04+
```
0xTim, line 40: This should be 20.04, 18.04 is almost EoL

Sources/SQLKit/Builders/SQLAlterTableBuilder.swift
```diff
@@ -1,290 +1,158 @@
...
+    /// Create a new ``SQLAlterTableBuilder``.
+    @inlinable
+    public init(_ alterTable: SQLAlterTable, on database: any SQLDatabase) {
         self.alterTable = alterTable
         self.database = database
     }
     
-    @discardableResult
     /// Rename the table.
-    /// - Parameter newName: The new name to apply to the table
```
0xTim, line 24: What's with removing the parameters from the docs? These would be good for DocC right?

Sources/SQLKit/Concurrency/SQLQueryFetcher+Concurrency.swift
```diff
@@ -2,26 +2,40 @@ import NIOCore
...
     }
     
-    func all() async throws -> [SQLRow] {
-        return try await self.all().get()
+    func all() async throws -> [any SQLRow] {
+        try await self.all().get()
     }
     
+    #if swift(>=5.7)
+    @preconcurrency
```
0xTim, line 21: Why do we need this here?

Sources/SQLKit/Query/SQLDataType.swift
```diff
@@ -6,34 +6,28 @@ public enum SQLDataType: SQLExpression {
     case text
     case real
     case blob
+    case custom(any SQLExpression)
```
0xTim, line 9: How breaking is this

0xTim (approved):
Code wise looks good. A few questions - main concern is around the removal of parameters and return types in the comments which would have been used by DocC

0xTim:
Oh one other comment I had is that with the full readme, DocC support etc this is now ready for a SSWG pitch

Sources/SQLKit/Builders/SQLAlterTableBuilder.swift
gwynne: In basically every case, they're pointless. "The name is the name" etc.; it's just absurd.

Sources/SQLKit/Concurrency/SQLQueryFetcher+Concurrency.swift
gwynne: Truthfully? I don't actually know, I've just been mimicking what NIO does when they add `@Sendable` to closure parameters.

Sources/SQLKit/Query/SQLDataType.swift
gwynne: It isn't, it's just moved from further down in the file.

gwynne:
> Oh one other comment I had is that with the full readme, DocC support etc this is now ready for a SSWG pitch

Strongly disagree. There are major design flaws with SQLKit despite this cleanup, I'd much rather wait for what I'll be doing with it for Fluent 5.

Sources/SQLKit/Concurrency/SQLQueryFetcher+Concurrency.swift
0xTim: I'm pretty sure we can remove these instances then. They're mainly used to silence warnings and I can't see anything here that would trigger a warning

0xTim: I'm pretty sure we can remove these instances then. They're mainly used to silence warnings and I can't see anything here that would trigger a warning

Sources/SQLKit/Builders/SQLAlterTableBuilder.swift
0xTim: Ok let's leave it, we can always backfill if necessary

PR #171, opened 2023-12-10
Add Model Macro for `select`, `insert`, `drop`

zunda-pixel:
## Adds a new `Model` attribute.

`Model` attribute is Swift Macro.
`Model` attribute adds `columnsName: [String]`, `values: [any Encodable]`, `Modelable`, `Decodable`.

## Example

```swift
@Model
struct User {
  var firstName: String
  var lastName: String
    
  var fullName: String { "\(fullName) \(lastName)" }
      
  var age: Int

  @ModelableIgnored
  var other: String
}
```

## Inline Macro

```swift
struct User {
  var firstName: String
  var lastName: String
  var fullName: String { "\\(fullName) \\(lastName)" }
  
  var age: Int

  @ModelableIgnored
  var other: String

  static public var columnNames: [String] {
    ["firstName", "lastName", "age"]
  }

  public var values: [any Encodable] {
    [self.firstName, self.lastName, self.age]
  }
}

extension User : Modelable, Decodable {
}
```


0xTim:
Hi @zunda-pixel, thank you for your submission. Unfortunately we're not able to accept this pull request. There are a number of issues with it, the main ones being that this additional functionality is out of scope for what SQLKit aims to achieve and the `@Model` macro is going to cause significant issues with Fluent's `Model` protocol (not to mentioned we still support Swift 5.7 and 5.8).

Any significant changes like this need to be discussed first so we can flesh out scope, API and functionality. The best option is to create an issue or post in Vapor's `#development` or `#fluent-5` channels on Discord (since this kind of macro is something that would be better suited to FluentKit and something we already have planned).

PR #172, opened 2024-03-20
Major overhaul of the entire SQLKit package

gwynne:
This will be the last release of SQLKit version 3.

A hopefully complete (but probably not) list of significant changes:

- A massive sweep to add at least minimal documentation to everything in the package.
- 100% test coverage.
- Reorganized the source code layout
- Require Swift 5.8
- Package is now `ExistentialAny`-compliant
- Numerous poorly-designed APIs have been deprecated. Replacement suggestions are available in all cases.
- 100% `Sendable`-complaint (zero concurrency warnings).
- `SQLDatabaseReportedVersion` is now `Equatable` and `Comparable`, as it ought to have been from the start.
- Efficiency improvements for the `async` versions of various APIs.
- `SQLQueryEncoder` and `SQLRowDecoder` have been massively overhauled; both are now considerably more flexible and less restrictive.
- `SQLBenchmark` is now `async`.
- Numerous ugly `assert()`s and `print()`s are now consistently routed through the database's logger instead, and less noisy logging is done.
- Several bits of missing functionality in `SQLCreateTrigger` are now correctly implemented.
- `SQLIdentifier` and `SQLLiteral.string` now automatically escape the appropriate quote characters when serializing.
- Added `SQLBetween` (`x BETWEEN y AND z`), `SQLQualifiedTable` (`schema.table`), `SQLSubquery` and `SQLSubqueryBuilder`, `SQLUnqualifiedColumnListBuilder`, and `SQLAliasedColumnListBuilder`.
- `SQLPredicateBuilder` and `SQLSecondaryPredicateBuilder` now provide 1-to-1 corresponding APIs for all four variants ("and where", "and having", "or where", "or having").
- Serialization of expressions is now a bit faster across the board.
- `SQLInsert` and `SQLInsertBuilder` now support the `INSERT ... SELECT` syntax.
- `SQLQueryFetcher` gained a number of convenience APIs for decoding models and single columns. Several builders also gained convenience methods for encoding or decoding of models. (Note that "models" in this usage does _NOT_ refer to Fluent's `Model` protocol, but rather to any `Codable` structure.)
- `SQLDropBehavior` is now used by all builders that support the modifier and respects the current dialect properly.
- `SQLCreateIndex` now supports index predicates.
- `SQLDistinct` now serializes correctly.
- Several expressions and queries are now more tolerant of missing configuration when serialized.
- `SQLDatabase` gained a `withSession(_:)` API which, when implemented correctly by drivers, allows implementing transactions safely using pure SQLKit (no need for Fluent-level or driver-level intervention).
- `SQLColumnConstraintAlgorithm.primaryKey` no longer emits incorrect SQL if the active dialect uses `NULL` as its `DEFAULT` literal.


Sources/SQLKit/Builders/Implementations/SQLInsertBuilder.swift
```diff
@@ -0,0 +1,384 @@
+/// Builds ``SQLInsert`` queries.
+///
+/// > Note: Although in the strictest sense, this builder could conform to ``SQLUnqualifiedColumnListBuilder``, doing
+/// > so would be semantically inappropriate. the protocol documents its `columns()` methods as being additive, but
```
ptoffy, line 4:
```suggestion
/// > so would be semantically inappropriate. The protocol documents its `columns()` methods as being additive, but
```

Sources/SQLKit/Builders/Implementations/SQLInsertBuilder.swift
```diff
@@ -0,0 +1,384 @@
...
+    ///     .run()
+    ///
+    /// // Effectively the same as:
+    /// try await sqlDatabase.insert(into: "planets")
+    ///     .columns("id", "name", "is_inhabited")
+    ///     .values(SQLBind(earth.id), SQLBind(earth.name), SQLBind(earth.isInhabited))
+    ///     .run()
+    /// ```
+    ///
+    /// > Note: The term "model" does _not_ refer to Fluent's `Model` type. Fluent models are not compatible with
```
ptoffy, line 53: IMO this isn't needed, this package can be used without Fluent too so people who might do that could be confused

Sources/SQLKit/Builders/Implementations/SQLInsertBuilder.swift
```diff
@@ -0,0 +1,384 @@
...
+    ///     .run()
+    ///
+    /// // Effectively the same as:
+    /// try await sqlDatabase.insert(into: "planets")
+    ///     .columns("id", "name", "isInhabited")
+    ///     .values(SQLLiteral.null, SQLBind(earth.name), SQLBind(earth.isInhabited))
+    ///     .run()
+    /// ```
+    ///
+    /// > Note: The term "model" does _not_ refer to Fluent's `Model` type. Fluent models are not compatible with
```
ptoffy, line 99: IMO this isn't needed, this package can be used without Fluent too so people who might do that could be confused

Sources/SQLKit/Builders/Implementations/SQLInsertBuilder.swift
```diff
@@ -0,0 +1,384 @@
...
+    ///
+    /// // Effectively the same as:
+    /// try await sqlDatabase.insert(into: "planets")
+    ///     .columns("id", "name", "is_inhabited")
+    ///     .values(SQLBind(earth.id), SQLBind(earth.name), SQLBind(earth.isInhabited))
+    ///     .values(SQLBind(mars.id), SQLBind(mars.name), SQLBind(mars.isInhabited))
+    ///     .run()
+    /// ```
+    ///
+    /// > Note: The term "model" does _not_ refer to Fluent's `Model` type. Fluent models are not compatible with
```
ptoffy, line 134: IMO this isn't needed, this package can be used without Fluent too so people who might do that could be confused

Sources/SQLKit/Builders/Implementations/SQLInsertBuilder.swift
```diff
@@ -0,0 +1,384 @@
...
+    ///
+    /// // Effectively the same as:
+    /// try await sqlDatabase.insert(into: "planets")
+    ///     .columns("id", "name", "isInhabited")
+    ///     .values(SQLLiteral.null, SQLBind(earth.name), SQLBind(earth.isInhabited))
+    ///     .values(SQLLiteral.null, SQLBind(mars.name), SQLBind(mars.isInhabited))
+    ///     .run()
+    /// ```
+    ///
+    /// > Note: The term "model" does _not_ refer to Fluent's `Model` type. Fluent models are not compatible with
```
ptoffy, line 176: IMO this isn't needed, this package can be used without Fluent too so people who might do that could be confused

Sources/SQLKit/Database/SQLDatabase.swift
```diff
@@ -0,0 +1,265 @@
...
+    var dialect: any SQLDialect {
+        self.database.dialect
+    }
+
+    // See `SQLDatabase.queryLogLevel`.
+    var queryLogLevel: Logger.Level? {
+        self.database.queryLogLevel
+    }
+    
+    // See `SQLDatabase.execut(sql:_:)`.
```
ptoffy, line 250:
```suggestion
    // See `SQLDatabase.execute(sql:_:)`.
```

Sources/SQLKit/Database/SQLDatabase.swift
```diff
@@ -0,0 +1,265 @@
...
+    func execute(
+        sql query: any SQLExpression,
+        _ onRow: @escaping @Sendable (any SQLRow) -> ()
+    ) -> EventLoopFuture<Void> {
+        self.database.execute(sql: query, onRow)
+    }
+
+    // See `SQLDatabase.execut(sql:_:)`.
```
ptoffy, line 258:
```suggestion
    // See `SQLDatabase.execute(sql:_:)`.
```

Sources/SQLKit/Database/SQLDatabaseReportedVersion.swift
```diff
@@ -0,0 +1,99 @@
+/// Provides a protocol for reporting and comparing database version numbers.
+///
+/// SQLKit allows databases to report their versions. As any given database implementation
+/// may have its own particular format for version numbers, the version is provided to
+/// ``SQLDatabase`` as a value of a type conforming to this protocol, which defines an
+/// interface for generically handling and comparing versions without needing to be aware
+/// of implementation-specific details.
+///
+/// The most common uses for database version information are disabling or enabling feature
+/// suport in ``SQLDialect`` by version, tracking usage metrics by version, logging versions,
```
ptoffy, line 10:
```suggestion
/// support in ``SQLDialect`` by version, tracking usage metrics by version, logging versions,
```

Sources/SQLKit/Database/SQLDatabaseReportedVersion.swift
```diff
@@ -0,0 +1,99 @@
...
+/// and recording versions for debugging.
+///
+/// Each type implementing ``SQLDatabaseReportedVersion`` is responsible for providing
+/// defintions of equality and ordering semantics between versions which are meaningful
```
ptoffy, line 14:
```suggestion
/// definitions of equality and ordering semantics between versions which are meaningful
```

Sources/SQLKit/Deprecated/SQLExoressions+Deprecated.swift
```diff
@@ -0,0 +1,125 @@
...
+/// Old name for ``SQLCreateTrigger/EachSpecifier``.
+@available(*, deprecated, renamed: "SQLCreateTrigger.EachSpecifier")
+public typealias SQLTriggerEach = SQLCreateTrigger.EachSpecifier
+
+/// Old name for ``SQLCreateTrigger/OrderSpecifier``.
+@available(*, deprecated, renamed: "SQLCreateTrigger.OrderSpecifier")
+public typealias SQLTriggerOrder = SQLCreateTrigger.OrderSpecifier
+
+extension SQLUnionJoiner {
+    @available(*, deprecated, message: "Use .type` instead.")
```
ptoffy, line 105:
```suggestion
    @available(*, deprecated, message: "Use `.type` instead.")
```

Sources/SQLKit/Deprecated/SQLExoressions+Deprecated.swift
```diff
@@ -0,0 +1,125 @@
...
+            case (.unionAll, false): self.type = .union
+            case (.intersect, true): self.type = .intersectAll
+            case (.intersectAll, false): self.type = .intersect
+            case (.except, true): self.type = .exceptAll
+            case (.exceptAll, false): self.type = .except
+            default: break
+        } }
+    }
+    
+    @available(*, deprecated, message: "Use .init(type:)` instead.")
```
ptoffy, line 120:
```suggestion
    @available(*, deprecated, message: "Use `.init(type:)` instead.")
```

Sources/SQLKit/Docs.docc/SQLKit.md
```diff
@@ -0,0 +1,385 @@
+# ``SQLKit``
+
+@Metadata {
+    @TitleHeading(Package)
+}
+
+SQLKit is an library for building and serializing SQL queries in Swift.
```
ptoffy, line 7:
```suggestion
SQLKit is a library for building and serializing SQL queries in Swift.
```

Sources/SQLKit/Docs.docc/SQLKit.md
```diff
@@ -0,0 +1,385 @@
...
+Instances of ``SQLDatabase`` are capable of serializing and executing ``SQLExpression``s:
+
+```swift
+let db: any SQLDatabase = // ...
+db.execute(sql: db.serialize(db.select().column(SQLLiteral.string("a")).query), onRow: { (row: any SQLRow) in
+    // ...
+})
+```
+
+The ``SQLExpression`` protocol provides a common interface for transforming an arbitrary set of syntactical building blocks into a string of SQL code. A comprehensive set of SQL building blocks for SQL syntax   various syntactical building blocks  which abstracts an arbitrary sequence of that represents a SQL query string and optional bind values. It can represent an entire SQL query or just a fragment.
```
ptoffy, line 168: This phrase seems incomplete

Sources/SQLKit/Docs.docc/SQLKit.md
```diff
@@ -0,0 +1,385 @@
...
+
+### Rows
+
+For query builders that support returning results (e.g. any builder conforming to the `SQLQueryFetcher` protocol), there are additional methods for handling the database output:
+
+- `all()`: Returns an array of rows.
+- `first()`: Returns an optional row.
+- `run(_:)`: Accepts a closure that handles rows as they are returned.
+
+Each of these methods returns `SQLRow`, which has methods for access column values.
```
ptoffy, line 211:
```suggestion
Each of these methods returns `SQLRow`, which has methods for accessing column values.
```

Sources/SQLKit/Docs.docc/SQLKit.md
```diff
@@ -0,0 +1,385 @@
+# ``SQLKit``
+
+@Metadata {
+    @TitleHeading(Package)
+}
+
+SQLKit is an library for building and serializing SQL queries in Swift.
+
+SQLKit's query construction facilities provide mappings between Swift types and database field types, and a direct interface to executing SQL queries. It attempts to abstract away the many differences between the various dialects of SQL whenever practical, allowing users to construct queries for use with any of the supported database systems. Custom SQL can be directly specified as needed, such as when abstraction of syntax is not possible or unimplemented.
```
ptoffy, line 9:
```suggestion
SQLKit's query construction facilities provide mappings between Swift types and database field types, and a direct interface for executing SQL queries. It attempts to abstract away the many differences between the various dialects of SQL whenever practical, allowing users to construct queries for use with any of the supported database systems. Custom SQL can be directly specified as needed, such as when abstraction of syntax is not possible or unimplemented.
```

Sources/SQLKit/Expressions/Basics/SQLDirection.swift
```diff
@@ -0,0 +1,29 @@
+/// Describes a ordering direction for a given sorting key.
```
ptoffy, line 1:
```suggestion
/// Describes an ordering direction for a given sorting key.
```

Sources/SQLKit/Expressions/Basics/SQLDirection.swift
```diff
@@ -0,0 +1,29 @@
...
+public enum SQLDirection: SQLExpression {
+    /// Ascending order (minumum to maximum), as defined by the sorting key's data type.
```
ptoffy, line 3:
```suggestion
    /// Ascending order (minimum to maximum), as defined by the sorting key's data type.
```

Sources/SQLKit/Expressions/Basics/SQLQueryString.swift
```diff
@@ -0,0 +1,267 @@
+/// An expression consisting of an array of constituent subexpressions generated by custom string interpolations.
+///
+/// Query strings are primarily intended for use with ``SQLRawBuilder``, providing for the inclusion of bound
+/// parameters in otherwise "raw" queries. The API alsos support some of the more commonly used quoting functionality.
```
ptoffy, line 4:
```suggestion
/// parameters in otherwise "raw" queries. The API also supports some of the more commonly used quoting functionality.
```

ptoffy (changes requested):
Mostly nits around comments, text etc.

Sources/SQLKit/Builders/Implementations/SQLInsertBuilder.swift
gwynne: In this version of the package, Fluent is tied to its usage in practically all use cases, so I think the warning is appropriate. However, this will not be the case after this point, so these can be removed in the major version update.

Sources/SQLKit/Builders/Implementations/SQLAlterTableBuilder.swift
```diff
@@ -3,22 +3,15 @@ public final class SQLAlterTableBuilder: SQLQueryBuilder {
...
-    /// See ``SQLQueryBuilder/query``.
+    // See `SQLQueryBuilder.query`.
     @inlinable
     public var query: any SQLExpression {
         self.alterTable
     }
 
-    /// The set of column alteration expressions.
-    @inlinable
-    public var columns: [any SQLExpression] {
```
0xTim, line 17: I thought this PR wasn't a breaking change?

0xTim (approved):
Overall I think this LGTM. I have a few comments about API breakages, but I'm not sure if they're true or not.

Great to see lots of extra tests and documentation coverage. I will say this PR is very hard to review due to the sheer size and changes involved, it probably would have been better to break it up into separate PRs as there's definitely places where things could be missed

Sources/SQLKit/Expressions/Queries/SQLCreateTrigger.swift
```diff
@@ -58,133 +124,179 @@ public struct SQLCreateTrigger: SQLExpression {
...
         let syntax = serializer.dialect.triggerSyntax.create
-        let when = self.when as? SQLTriggerWhen, event = self.event as? SQLTriggerEvent, each = self.each as? SQLTriggerEach
+        let when = self.when as? WhenSpecifier, event = self.event as? EventSpecifier, each = self.each as? EachSpecifier
         
         if syntax.contains(.postgreSQLChecks) {
-            assert(when != .instead || event != .update || self.columns == nil, "INSTEAD OF UPDATE events do not support lists of columns")
-            assert(when != .instead || each == .row, "INSTEAD OF triggers must be FOR EACH ROW")
-            assert(!syntax.contains(.supportsUpdateColumns) || (columns?.isEmpty ?? true) || event != .update, "Only UPDATE triggers may specify a list of columns.")
-            assert(!syntax.contains(.supportsCondition) || when != .instead || self.condition == nil, "INSTEAD OF triggers do not support WHEN conditions.")
+            if !(when != .instead || event != .update || (self.columns?.isEmpty ?? true)) { serializer.database.logger.warning("INSTEAD OF UPDATE events do not support lists of columns") }
```
0xTim, line 146: Warning logs again

Sources/SQLKit/Expressions/Queries/SQLCreateTrigger.swift
```diff
@@ -58,133 +124,179 @@ public struct SQLCreateTrigger: SQLExpression {
...
-                $0.append("END;")
+                $0.append("BEGIN", SQLList(body, separator: SQLRaw(" ")), "END;")
             } else if let procedure = self.procedure {
                 $0.append("EXECUTE PROCEDURE", procedure)
             }
         }
     }
 }
 
-public enum SQLTriggerWhen: SQLExpression {
```
0xTim, line 117: Is this breaking?

Sources/SQLKit/Expressions/Queries/SQLDropEnum.swift
```diff
@@ -1,39 +1,46 @@
...
     public var ifExists: Bool
 
-    /// The optional `CASCADE` clause drops other objects that depend on this type
-    /// (such as table columns, functions, and operators), and in turn all objects
-    /// that depend on those objects.
-    public var cascade: Bool
+    /// A drop behavior.
+    ///
+    /// Ignored if not supported by the dialect. See ``SQLDropBehavior``.
+    public var dropBehavior: SQLDropBehavior
```
0xTim, line 23: We happy with this break?

Sources/SQLKit/Expressions/Queries/SQLDropTrigger.swift
```diff
@@ -1,44 +1,50 @@
...
     public var ifExists = false
 
-    /// The optional `CASCADE` clause drops other objects that depend on this type
-    /// (such as table columns, functions, and operators), and in turn all objects
-    /// that depend on those objects.
-    public var cascade = false
+    /// A drop behavior.
+    ///
+    /// Ignored if not supported by the dialect. See ``SQLDropBehavior``.
+    public var dropBehavior = SQLDropBehavior.restrict
```
0xTim, line 25: Again happy with the break?

Sources/SQLKit/Expressions/SQLStatement.swift
```diff
@@ -7,62 +53,190 @@ extension SQLSerializer {
...
+        iter.next()?.serialize(to: &serializer)
+        while let part = iter.next() {
+            var temp = SQLSerializer(database: serializer.database)
+            temp.binds = serializer.binds
+            
+            part.serialize(to: &temp)
+            if !temp.sql.isEmpty {
+                serializer.sql += " \(temp.sql)"
+            }
+            serializer.binds = temp.binds//.append(contentsOf: temp.binds) // Can't just append because we need to keep numbers in sync
```
0xTim, line 125: Do we want to remove the comment out code?

Sources/SQLKit/Expressions/Syntax/SQLBinaryOperator.swift
```diff
@@ -1,94 +1,95 @@
...
+        case .add:                serializer.write("+")
+        case .subtract:           serializer.write("-")
 
         // See https://dev.mysql.com/doc/refman/8.0/en/sql-mode.html#sqlmode_pipes_as_concat
         case .concatenate:
-            fatalError("""
-                || is not implemented because MySQL doesn't always support it, even though everyone else does.
-                Use `SQLFunction("CONCAT", args...)` for MySQL or `SQLRaw("||")` with Postgres and SQLite.
-                """)
+            serializer.database.logger.warning("|| is not implemented, because it doesn't always work. Use `SQLFunction(\"CONCAT\", args...)` for MySQL or `SQLRaw(\"||\")` for Postgres and SQLite.")
```
0xTim, line 92: Warning again

0xTim (approved):
Some queries, but looks good to go for the most part

Sources/SQLKit/Expressions/Clauses/SQLOrderBy.swift
```diff
@@ -1,20 +1,36 @@
...
+    // See `SQLExpression.serialize(to:)`.
     @inlinable
     public func serialize(to serializer: inout SQLSerializer) {
-        self.expression.serialize(to: &serializer)
-        serializer.write(" ")
-        self.direction.serialize(to: &serializer)
+        /// The mere fact that the serialization of this clause is this trivial underscores how excessively verbose
+        /// making use of it is and why it is superfluous in a better-designed API.
+        serializer.statement {
+            $0.append(self.expression, self.direction)
```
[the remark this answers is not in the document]
gwynne: Yep. But not usable in 100% of cases, since statement appends always include whitespace between each item; there are a few places where I still use the serializer writes directly to avoid that.

Sources/SQLKit/Expressions/Clauses/SQLReturning.swift
```diff
@@ -1,30 +1,45 @@
...
     public init(_ columns: [any SQLExpression]) {
         self.columns = columns
     }
 
+    // See `SQLExpression.serialize(to:)`.
     public func serialize(to serializer: inout SQLSerializer) {
         guard serializer.dialect.supportsReturning else {
-            serializer.database.logger.warning("\(serializer.dialect.name) does not support 'RETURNING' clause, skipping.")
+            /// This logging can be a bit noisy for MySQL users, elide it for now.
+            //serializer.database.logger.warning("\(serializer.dialect.name) does not support 'RETURNING' clause, skipping.")
```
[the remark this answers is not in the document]
gwynne: I have been thoroughly inconsistent in my handling of places where there were assertions, logs, etc. in the existing code, but I've tried to tend towards silencing things completely (since in the present day it generally isn't useful to have the warnings anyway). The much more correct solution would be for serialization to be able to throw so that obvious syntax problems can be reported properly, but of course that requires the upcoming major version break.

Sources/SQLKit/Expressions/Queries/SQLAlterTable.swift
```diff
@@ -1,73 +1,83 @@
...
-
-        let removals = (self.dropColumns + self.dropTableConstraints).map { column in
-            (verb: SQLRaw("DROP"), definition: column)
-        }
-
-        let alterColumnDefinitionCaluse = syntax.alterColumnDefinitionClause ?? SQLRaw("MODIFY")
-        let modifications = self.modifyColumns.map { column in
-            (verb: alterColumnDefinitionCaluse, definition: column)
+        if syntax.alterColumnDefinitionClause == nil, !self.modifyColumns.isEmpty {
+            serializer.database.logger.warning("Database does not support column modifications.")
```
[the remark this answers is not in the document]
gwynne: My justification for using `warning` level is that the logging here (and in the other places I do it) is reporting a major functional concern, i.e. "this code WILL NOT DO what was intended", and in particular in an area where the failure to behave as intended 1) may not be obvious at runtime in either debug or production, and 2) is likely to have serious data integrity and/or security implications.

Sources/SQLKit/Expressions/Queries/SQLCreateEnum.swift
```diff
@@ -1,25 +1,41 @@
...
+/// 
 /// See ``SQLCreateEnumBuilder``.
 public struct SQLCreateEnum: SQLExpression {
-    /// Name of type to create.
+    /// The name for the created type.
     public var name: any SQLExpression
 
+    /// The enumeration values for the new type.
+    ///
+    /// Must contain at least one value.
```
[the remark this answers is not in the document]
gwynne: We don't. There are NUMEROUS other places where an array must have at least one value but this is never asserted; it is left to the underlying execution of the query against the database to report the error (since the "must" in this context means "otherwise invalid SQL will be generated").

Sources/SQLKit/Expressions/Queries/SQLCreateTrigger.swift
gwynne: No; the name is typealiased over in `SQLExpressions+Deprecated.swift`

Sources/SQLKit/Expressions/Queries/SQLDropEnum.swift
gwynne: There is no breakage. No memberwise initializer is public (or exists, for that matter), and the default value of the additional property has the same effect as when it didn't exist.

Sources/SQLKit/Expressions/Queries/SQLDropTrigger.swift
gwynne: Same as before, no actual break exists.

Sources/SQLKit/Expressions/SQLStatement.swift
gwynne: I left it in for illustrative purposes, but I'm okay with removing it.

Sources/SQLKit/Expressions/Clauses/SQLReturning.swift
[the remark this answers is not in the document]
0xTim: How about an assertion to trigger in debug and then a log at debug as per the log level guidelines?

ptoffy (approved):

Sources/SQLKit/Expressions/Syntax/SQLRaw.swift
```diff
@@ -1,16 +1,23 @@
...
+    @inlinable
+    public init(_ sql: String, _ binds: [any Encodable]) {
         self.sql = sql
         self.binds = binds
     }
     
     @inlinable
     public func serialize(to serializer: inout SQLSerializer) {
         serializer.write(self.sql)
-        serializer.binds += self.binds
```
arennow, line 14: Just as a note, this removal makes this a (pretty big) breaking change. I'm not sure there's anything to be done about it now, since the release was, I believe, quite a while ago. But just a note for the future

gwynne: It was a very much deliberate breakage. This particular "functionality" of `SQLRaw` was a misfeature accidentally left in the code during Tanner's work on Vapor 4; it should never have been present to begin with. I couldn't remove the `binds` property itself without causing source breakage, but I could at least discourage its usage.

gwynne: _That being said..._

If you are a user whose code was broken by this change, I'm not dead-set against undoing it, even this long afterwards; putting it back would be a purely additive change. The `binds` property will be gone anyway in Fluent 5, so it doesn't matter as much for Fluent 4 at this point.

arennow: I don't personally need it (I only used it in two places, and I figured out the `\(bind: bindVal)` thing). I just think it's _unexpected_ for a minor version bump to remove significant (though apparently unintentional) runtime functionality with only a compile-time warning as an indication

Idk if this project uses Semver. If not, then the expectation that that wouldn't happen is my fault, and you should just ignore me :)

0xTim: Just to nitpick 😅  - SemVer definitely allows breaking behaviour changes, otherwise you'd never be able to fix bugs as that's a change in behaviour. SemVer is purely about API changes, and in Swift's case, whether something still compiles, not if it behaves the same way (not saying that's any comfort for having something major changed and we tend to try and avoid that in most places)

PR #175, opened 2024-04-15
fix: Adds back-support for SQLBind public API

NeedleInAJayStack:
This restores the public API of `SQLBind`. It appears that the `any Encodable` argument support was dropped between v3.28.0 and v3.29.0.

Please let me know if you'd like these functions marked as `Deprecated` and placed in the `Deprecated` source directory.

Thanks for all your work on this package!!

### DIFF
--- a/Sources/SQLKit/Expressions/Syntax/SQLBind.swift
+++ b/Sources/SQLKit/Expressions/Syntax/SQLBind.swift
@@ -8,10 +8,23 @@ public struct SQLBind: SQLExpression {
     public init(_ encodable: some Encodable & Sendable) {
         self.encodable = encodable
     }
+
+    /// Create a binding to a value.
+    @inlinable
+    @_disfavoredOverload
+    public init(_ encodable: any Encodable & Sendable) {
+        self.encodable = encodable
+    }
     
     /// Create a list of bindings to an array of values, with the placeholders wrapped in an ``SQLGroupExpression``.
     @inlinable
     public static func group(_ items: some Collection<some Encodable & Sendable>) -> any SQLExpression {
+        SQLGroupExpression(items.map(SQLBind.init))
+    }
+
+    /// Create a list of bindings to an array of values, with the placeholders wrapped in an ``SQLGroupExpression``.
+    @inlinable
+    public static func group(_ items: [any Encodable & Sendable]) -> any SQLExpression {
         SQLGroupExpression(items.map(SQLBind.init))
     }
 


gwynne:
The `some Encodable & Sendable` API should have been a source-compatible replacement for `any Encodable & Sendable`, are you finding otherwise?

NeedleInAJayStack:
> are you finding otherwise?

Yeah, unfortunately. We have a variable that holds an `any Encodable` and uses it with SQLBind. When using with the new API, it fails to compile with the following error:

```
error: type 'any Encodable' cannot conform to 'Encodable'
```

My understanding is that `some` can be converted to `any`, but `any` cannot be converted to `some`.

gwynne:
Are you by any chance using Swift 5.8? I've been able to verify that the existential is automatically unwrapped when using 5.9 or 5.10.

NeedleInAJayStack:
Hmm, weird - no I'm using Swift 5.10. Let me see if I can get an easy reproducer.

gwynne:
Okay, based on some further poking around into the compiler's behavior, I think this may be a case of this is _supposed_ to work but doesn't in some contexts (for example, I can reproduce the error you're seeing if I make the property an IUO, e.g. `(any Encodable & Sendable)!`).

NeedleInAJayStack:
Ah yes, I guess it was that we were passing in `[any Encodable & Sendable]` which ends up similar to your `(any Encodable & Sendable)!` example. Simple reproducer:

```swift
import SQLKit

func reproduce(a: [any Encodable & Sendable]) {
  let sqlBind = SQLBind.group(a) // error: type 'any Encodable & Sendable' cannot conform to 'Encodable'
}
```

gwynne:
Oof, yeah, that won't work. Let me look a little deeper into this. I thought I'd avoided the `some`/`any` pitfalls, but I obviously missed at least a couple; there are probably others.

gwynne:
Okay yeah, the entire `SQLPredicateBuilder` API has this problem too, no wonder I was using `@_disfavoredOverload` in the old version... I'll need to do a larger PR to get rid of the source-compat breakage. Guess type cleanliness will have to wait for the semver-major bump 😕

NeedleInAJayStack:
Oh dang. Okay, well thanks for your hard work on this package!! Feel free to close this MR if you loop the effort into a larger fix.

gwynne:
Thanks for reporting this! I probably would never have noticed it myself 😅

gwynne:
Closing in favor of #176, thanks again for making this known!

PR #176, opened 2024-04-16
Fix overhaul breakage, part 2

gwynne:
This solves the source code breakage issue first reported in #175 - shout out and thanks to @NeedleInAJayStack for reporting the problem!

Several preexisting APIs had incorrectly changed from accepting `any Encodable` to accepting `some Encodable`, which is source-breaking under some conditions. This restores the original use of `any` (though it keeps the added `Sendable` requirement). 

Also restores 100% test coverage after the previous fixes.

> [!NOTE]
> Many APIs which had previously accepted a generic parameter (i.e. `<E: Encodable>`), most notably in `SQLPredicateBuilder`, also switched to using `some Encodable`, but this was not source-breaking; the problem applied only to APIs which originally accepted `any Encodable`.
>
> Although the changes in this PR are technically themselves source-breaking, since they revert a previous such breakage to its previous state, only a `semver-patch` bump is necessary.

### DIFF
--- a/Sources/SQLKit/Builders/Prototypes/SQLColumnUpdateBuilder.swift
+++ b/Sources/SQLKit/Builders/Prototypes/SQLColumnUpdateBuilder.swift
@@ -76,7 +76,7 @@ extension SQLColumnUpdateBuilder {
     ///   - bind: The value to assign to the named column.
     @inlinable
     @discardableResult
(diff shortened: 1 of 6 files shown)


MahdiBM (approved):
Oof, didn't know `any` to `some` conversion doesn't work in these situations either.

NeedleInAJayStack:
Thanks @gwynne!!

PR #178, opened 2024-05-17
Support the use of unions in subqueries

gwynne:
Adds support for the use of `UNION` queries within subqueries. Unfortunately, thanks to iffy design choices on my part in the original `SQLUnion` implementation, the usage is slightly awkward. Example usage:

```swift
try await db.update("foos")
    .set(SQLIdentifier("bar_id"), to: SQLSubquery
        .union { $0
            .column("id")
            .from("bars")
            .where("baz", .notEqual, "bamf")
        }
        .union(all: { $0
            .column("id")
            .from("bars")
            .where("baz", .equal, "bop")
        })
        .finish()
    )
    .run()
```
This generates the following query:
```sql
UPDATE "foos"
SET "bar_id" = (
        SELECT "id" FROM "bars" WHERE "baz" <> "bamf"
    UNION ALL
        SELECT "id" FROM "bars" WHERE "baz" = "bop"
)
```

Unfortunately, it is not possible to chain `.union()` when using `SQLSubquery.select(_:)`; the call chain must start with `SQLSubquery.union(_:)`.


gwynne:
API breakage notes:

- The change from `SQLSelectBuilder` to `any SQLSubqueryClauseBuilder` in the parameters for the `SQSelectBuilder` extension methods is an API break but has no actual effect, as `SQLSelectBuilder` provides  no API that the protocol doesn't (or at least, none that can be used in the relevant context).
- The removed methods on `SQLUnionBuilder` are now provided as extension methods on the `SQLCommonUnionBuilder` protocol instead.

Sources/SQLKit/Builders/Prototypes/SQLCommonUnionBuilder.swift
```diff
@@ -0,0 +1,134 @@
...
+        self.union.add(query, joiner: .init(type: .intersect))
+       return self
+    }
+
+    /// Add a query to the union in `INTERSECT ALL` mode
+    /// (only results that come from both queries are returned, with duplicates preserved).
+    @inlinable
+    public func intersect(all query: SQLSelect) -> Self {
+        self.union.add(query, joiner: .init(type: .intersectAll))
+       return self
```
MahdiBM, line 61: Broken Indentations around here

gwynne: Fixed 🤦‍♀️

MahdiBM (approved):
If you're sure about the API breakage.

gwynne:
I am, actually 🙂. Look at the definition of `SQLSelectBuilder`. The only method of its own it defines is an initializer, and the `SQLQueryBuilder` and `SQLQueryFetcher` methods aren't callable within sub-builder closures (they're all asynchronous; there's nowhere to go with an ELF (anyone calling `.wait()` in that context frankly deserves what they get) and you can't `await` in a non-`async` closure).

PR #184, opened 2024-07-03
Add failing test for regression

stephencelis:
We have a project that started failing row decoding when upgrading from 3.28.0. I haven't dug into the problem but was able to sketch out a failing test that I believe should pass in 3.28.0 (wasn't able to easily confirm because the testing infrastructure on `main` didn't exist back then).

### DIFF
--- a/Tests/SQLKitTests/SQLRowDecoderTests.swift
+++ b/Tests/SQLKitTests/SQLRowDecoderTests.swift
@@ -172,6 +172,17 @@ final class SQLRowDecoderTests: XCTestCase {
             XCTAssertEqual(Array<any CodingKey>().map(\.stringValue), context.codingPath.map(\.stringValue))
         }
     }
+    
+    func testSnakeCase_CodableKeyTransformation() throws {
+        struct Episode: Decodable, Equatable {
+            var userID: Int
+        }
+
+        XCTAssertEqual(
+            try SQLRowDecoder(keyDecodingStrategy: .convertFromSnakeCase).decode(Episode.self, from: TestRow(data: ["user_id": "1"])),
+            Episode(userID: 1)
+        )
+    }
 }
 
 enum TestDecEnum: Codable {


gwynne:
I need to look at this in more detail (now that I'm back on the grid) - I seem to remember changing this because the previous behavior didn't match either the documentation or behavior of JSONEncoder/JSONDecoder (on which the strategy options were originally modeled). But it's been long enough that I'm not entirely sure that was actually the case, I may have just flubbed it when I rewrote the relevant logic.

gwynne:
Nope, I was right the first time, I fixed it to match `JSONDecoder`'s behavior:
```swift
Welcome to Swift version 6.0 (swift-6.0-RELEASE).
Type :help for assistance.
  1> import Foundation
  2> let d = JSONDecoder()
  3> d.keyDecodingStrategy = .convertFromSnakeCase
  4> struct Foo: Codable {
  5.     let userID: Int
  6. }
  7> try d.decode(Foo.self, from: Data("{\"user_id\":1}".utf8))
$E0: DecodingError = keyNotFound {
  keyNotFound = {
    0 = userID
    1 = {
      codingPath = 0 values
      debugDescription = "No value associated with key CodingKeys(stringValue: \"userID\", intValue: nil) (\"userID\")."
      underlyingError = nil
    }
  }
}
  8> struct Foo2: Codable {
  9.     let userId: Int
 10. }
 11> try d.decode(Foo2.self, from: Data("{\"user_id\":1}".utf8))
$R0: Foo2 = {
  userId = 1
}
```
(Note: Tested on both macOS and Linux with both Swift 5.10 and 6.0.)

gwynne:
@stephencelis I'm going to close this out as "works as expected", but if you have a counterargument, please feel free to reopen.

stephencelis:
@gwynne This behavior change is breaking existing code in our code base when we attempted to upgrade.

It's been awhile since I opened this and was familiar with the problem, but if I remember correctly I believe the core of the issue 2qw the snake case conversion changed:

```
Before:
  userID -> user_id

After:
  userID -> user_iD
```

stephencelis:
@gwynne OK rethinking about this I suppose it makes sense when it comes to roundtripping (though our app only cares about decoding at the moment). An unfortunate refactor awaits us some day :smile:

PR #185, opened 2024-08-03
Tests: Add missing `any` to a protocol composition

AnthonyLatsis:
The latest Swift release does not complain about this particular case because it was overlooked by `any` syntax diagnosis. Future compilers will.

See https://github.com/swiftlang/swift/issues/75655.


AnthonyLatsis:
cc @0xTim

0xTim (approved):

0xTim:
Thanks!

PR #189, opened 2025-04-16
Rename SQLRaw to SQLUnsafeRaw and SQLRawBuilder to SQLUnsafeRawBuilder

gwynne:
We really want to emphasize as much as possible that using raw SQL is unsafe. These changes also match up with the same renames in the Fluent 5 version of SQLKit. The original `SQLRaw` and `SQLRawBuilder` names are preserved as a deprecated `typealias`, as is the original `SQLDatabase.raw(_:)` spelling of what is now the `SQLDatabase.unsafeRaw(_:)` method.


0xTim (approved):
I have to say, something about this does feel off - essential all the unsafe usages inside _our_ codebase. I know it's a semantic thing and I don't really have a good answer other than adding internal functions for our own use. But I don't think it's an issue or a blocker, just something I noticed. (Also came up when thinking about other libraries such as FluentPostGIS - I guess we need to be able to opt into the unsafe things so this is probably the best way)

ptoffy (approved):

PR #190, opened 2025-06-25
Resolve issues breaking Swift Wasm builds for sql-kit

scottmarchant:
# Summary

This PR adds support for compiling sql-kit to wasm using the [Swift SDK for WebAssembly](https://www.swift.org/documentation/articles/wasm-getting-started.html).

This PR is [part of a larger effort](https://github.com/PassiveLogic/swift-web-examples/issues/1) by a company called PassiveLogic to enable broad support for Swift WebAssembly.

# Details

- Removed unused NIO dependency in Package.swift. This repo compiles fine without that, but having it breaks wasm compilation.
- Refined scope of exports to be `NIOCore` instead of the broader `NIO`.

# Notes

- Compiling to wasm requires specific versions of swift-nio. Because swift-nio still doesn't have wasm targets in it's CI, the wasm build breaks often. That issue [will be resolved soon](https://github.com/apple/swift-nio/pull/3159). Regardless, we don't want to bump the minimum nio requirement in the manifest, because that would be a breaking change for many using this repo. It is fine to leave the minimum nio requirement as-is. Developers trying to compile for wasm will almost certainly be aware of the issues compiling nio, and will soon have many wasm-compatible versions of nio to select.

# Testing done

- [x] Verified unit tests still pass with these changes
- [x] Verified `swift build` completes without errors
- [x] Verified no new compiler warnings are added with these changes
- [x] Verified `swift build --swift-sdk wasm32-unknown-wasi` completes without errors, when using a compatible nio version.
- [x] Verified `swift build --swift-sdk wasm32-unknown-wasip1-threads` completes without errors, when using a compatible nio version.
- [x] Verified a third-party executable can build this library as part of a larger wasm executable using the command `swift package --swift-sdk wasm32-unknown-wasip1-threads js --use-cdn`
- [x] Verified these changes compile in CI, using yet-to-be merged additions to vapor's CI. See See https://github.com/PassiveLogic/sql-kit/actions/runs/15886122859/job/44798410586?pr=1

# Impact Risk

There is a chance the changes to the manifest and exports could be a breaking change somewhere, if downstream dependencies relied on those pieces for their one and only intrinsic inclusion of the entire NIO dependency beyond NIOCore.

There are solutions to narrow the scope to just wasm compilation for those changes, if that is an issue. But I chose to leave them this way since it is a little cleaner solution.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,6 +1,21 @@
 // swift-tools-version:5.10
 import PackageDescription
 
+let nonWASIPlatforms: [Platform] = [
+    .macOS,
+    .macCatalyst,
+    .iOS,
+    .tvOS,
+    .watchOS,
+    .visionOS,
+    .driverKit,
+    .linux,
+    .windows,
+    .android,
+    // .wasi, // Everything but WASI
+    .openbsd,
+]
+
 let package = Package(
     name: "sql-kit",
     platforms: [
@@ -28,6 +43,7 @@ let package = Package(
             dependencies: [
                 .product(name: "Collections", package: "swift-collections"),
                 .product(name: "Logging", package: "swift-log"),
+                .product(name: "NIO", package: "swift-nio", condition: .when(platforms: nonWASIPlatforms)),
                 .product(name: "NIOCore", package: "swift-nio"),
             ],
             swiftSettings: swiftSettings
--- a/Package.swift
+++ b/Package.swift
@@ -16,9 +16,11 @@ let package = Package(
     dependencies: [
         .package(url: "https://github.com/apple/swift-collections.git", from: "1.1.0"),
         .package(url: "https://github.com/apple/swift-log.git", from: "1.5.4"),
-        // TODO: SM: Require a minimum swift-nio version once latest nio core fixes are versionized.
-        // .package(url: "https://github.com/apple/swift-nio.git", from: "2.65.0.TBD"),
-        .package(url: "https://github.com/apple/swift-nio.git", branch: "main"),
+        // NOTE: Compiling to wasm requires specific versions of swift-nio.
+        // This older version is left as-is to avoid placing restrictions
+        // on other targets. But to compile for wasm, make sure you have
+        // a version of swift-nio with a passing build for the NIOCore module.
+        .package(url: "https://github.com/apple/swift-nio.git", from: "2.65.0"),
     ],
     targets: [
         .target(
--- a/Package.swift
+++ b/Package.swift
@@ -31,11 +31,7 @@ let package = Package(
     dependencies: [
         .package(url: "https://github.com/apple/swift-collections.git", from: "1.1.0"),
         .package(url: "https://github.com/apple/swift-log.git", from: "1.5.4"),
-        // NOTE: Compiling to wasm requires specific versions of swift-nio.
-        // This older version is left as-is to avoid placing restrictions
-        // on other targets. But to compile for wasm, make sure you have
-        // a version of swift-nio with a passing build for the NIOCore module.
-        .package(url: "https://github.com/apple/swift-nio.git", from: "2.65.0"),
+        .package(url: "https://github.com/apple/swift-nio.git", from: "2.84.0"),
     ],
     targets: [
         .target(
--- a/Package.swift
+++ b/Package.swift
@@ -1,21 +1,6 @@
 // swift-tools-version:5.10
 import PackageDescription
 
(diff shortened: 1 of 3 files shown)


Package.swift
```diff
@@ -24,7 +28,6 @@ let package = Package(
             dependencies: [
                 .product(name: "Collections", package: "swift-collections"),
                 .product(name: "Logging", package: "swift-log"),
-                .product(name: "NIO", package: "swift-nio"),
```
scottmarchant, line 27: Note, there is a way to perform this elision for non-wasm builds only, if desired. This seemed like the best solution, though, since it appears that only NIOCore is being used currently in this repository.

Sources/SQLKit/Exports.swift
```diff
@@ -1,3 +1,3 @@
-@_documentation(visibility: internal) @_exported import protocol NIO.EventLoop
-@_documentation(visibility: internal) @_exported import class NIO.EventLoopFuture
```
scottmarchant, line 2: Note, there is a way to perform this elision for non-wasm builds only, if desired. This seemed like the best solution, though, since it appears that only NIOCore is being used currently in this repository.

Package.swift
```diff
@@ -16,6 +16,10 @@ let package = Package(
     dependencies: [
         .package(url: "https://github.com/apple/swift-collections.git", from: "1.1.0"),
         .package(url: "https://github.com/apple/swift-log.git", from: "1.5.4"),
+        // NOTE: Compiling to wasm requires specific versions of swift-nio.
```
0xTim, line 19: I'm happy to bump this to a version of NIO that supports Wasm. There's no issue bumping non-major versions of dependencies

scottmarchant: @0xTim My [fix for NIOCore](https://github.com/apple/swift-nio/pull/3271) is now in a NIO release, `2.84.0`. I think the most appropriate thing to do here per your comment is to bump the nio minor version.

> There's no issue bumping non-major versions of dependencies

If the [NIO wasm CI PR](https://github.com/apple/swift-nio/pull/3159) ever merges, we can make another pass to make that the new minimum.

scottmarchant: I bumped the minor NIO version.

Package.swift
scottmarchant: Per @gwynne 's comment below about fluent kit downstream problems, I decided to narrow the scope of this change to wasm-only platforms.

Sources/SQLKit/Exports.swift
scottmarchant: @gwynne That is a perfectly valid concern. I pushed a change that narrows preserves the old export and SPM dependency behavior, and makes the change only for WASI platforms. WASI cannot compile with the full `NIO.EventLoop` qualifier. It will only compile with `NIOCore.EventLoop`. So it wouldn't be possible for the latest changeset to break downstream WASI platform compilation, as they would be broken as-is without these changes.

And with the additional guards, non-WASI downstream dependencies should be 100% unaffected by this change, other than the bump in the minor NIO version.

Is there anything else you would like me to look into for this PR?

scottmarchant: @gwynne From what I see, all CI checks including the new wasm check are passing on this MR now.

gwynne (approved):

scottmarchant:
Note, if anyone encounters issues with this change due to reliance on the sql-kit exported of NIO, after consuming this change, you may need to add the following import if previously relying on the implicit import.

`import NIOCore`

Sources/SQLKit/Exports.swift
```diff
@@ -1,3 +1,8 @@
+#if os(WASI)
+@_documentation(visibility: internal) @_exported import protocol NIOCore.EventLoop
+@_documentation(visibility: internal) @_exported import class NIOCore.EventLoopFuture
```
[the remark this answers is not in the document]
scottmarchant: @gwynne As our company has started to consume this latest sql-kit change, we've found a few spots where consuming the latest sql-kit version in CI hits compiler errors. It tends to show up in places where we were only importing sql-kit and not anything from nio. I suspect the error may only happen on clean builds or something like that, because both you and I have contradictary experiences where this was non-breaking based on our own builds of the change.

We can leave this as is. It is not very hard for people to add new imports when consuming the change, if needed.

Or I'm happy to add the backwards compatibility for non-wasm targets if you'd prefer.

Any thoughts?

PR #202, opened 2026-07-30
feat: Add Wasm platform by conditionally building without swift-nio

scottmarchant:
Builds SQLKit on `wasm32-unknown-wasip1` without SwiftNIO. Since vapor/sql-kit#190 and swift-nio's WASI support, SQLKit itself already builds there with NIOCore linked; SwiftNIO is elided anyway because the drivers beneath SQLKit cannot build it at all (NIOPosix needs the POSIX sockets and threads WASI preview 1 lacks), so nothing on that platform can implement an `EventLoopFuture` surface, and keeping `canImport(NIOCore)` uniformly false across the stack means one condition selects the async-only configuration everywhere. Compiling under Embedded Swift needs a further set of gates, including a change to how the bound-parameter constraint is spelled; that work is kept out of this PR and is not being proposed yet, which keeps this diff small.

**Changes:**

- The legacy `EventLoopFuture` surface goes behind `#if canImport(NIOCore)`: `SQLDatabase.eventLoop`, the future-returning `execute(sql:_:)` requirement and its default bridge, `SQLQueryBuilder.run()`, the three `SQLQueryFetcher` future families, the NIOCore re-exports, and `SQLBenchmarker`'s two deprecated bridges. The `async` surface, already the recommended API at every one of those call sites, is untouched and is all that remains where SwiftNIO is absent.
- Each `#if` encloses the doc comment of the declaration it gates rather than sitting between the two. A `#if` in that position detaches the comment: the declaration still compiles, but the symbol graph reports it as undocumented and the published docs lose the entry, with no build-time diagnostic.
- The manifest gates `NIOCore` (and the test target's `NIOCore`/`NIOEmbedded`) with `.when(platforms: nonWASIPlatforms)`, the same spelled-out-list idiom as apple/swift-nio#2671 and swift-crypto, with a comment pinning the list to the manifest's tools version. The test suite's imports of those modules gate on the same condition, because `swift build` evaluates `--explicit-target-dependency-import-check` for every target in the graph, including the test target it does not build.

**Where SwiftNIO is present, nothing changes:** every gate is unconditionally true, and `swift package diagnose-api-breaking-changes` reports no breaking changes in SQLKit or SQLKitBenchmark. `swift build -Xswiftc -emit-symbol-graph` emits the same symbols with the same `docComment` line counts as before.

**What WASI loses:** the `EventLoopFuture` half of the API. `SQLDatabase.eventLoop`, the future-returning `execute(sql:_:)` and `SQLQueryBuilder.run()`, the future variants of the `SQLQueryFetcher` `first`/`all`/`run` families, the NIOCore re-exports, and `SQLBenchmarker`'s deprecated bridges. The `async` surface, already the recommended API, is what remains. One driver-facing nuance: the default implementation of the `async` `execute(sql:_:)` forwards to the future-returning requirement, so where SwiftNIO is absent there is no future overload to forward to and drivers implement the `async` requirement directly, which is the primitive they should be implementing anyway.

**Verification:**

- Native `swift build` and `swift test`: green, 169 tests, including with CI's `--explicit-target-dependency-import-check error -Xswiftc -require-explicit-sendable`.
- WASI (`--swift-sdk swift-6.3.1-RELEASE_wasm`): green with the same flags, zero SwiftNIO object files, no NIO symbols in the built objects. `SQLKitBenchmark` still builds for regular WASI.
- No CI change needed: sql-kit already passes `with_wasm: true` to `vapor/ci`.

**On tests:** no test cases are added. The change introduces no behavior on any platform that compiled before (every gate is unconditionally true where SwiftNIO is present, and the symbol graph is byte-identical), and the configuration it does add is exercised by the wasm CI lane.

**Notes for review:** every gate is a capability gate (`canImport`), in `Sources/` only, never in a manifest; there is no `os(WASI)` anywhere. All three packages in this stack gate the same modules on the same condition, and a package that did build NIOCore for WASI would fail loudly at compile. No SwiftPM traits, no versioned manifest, no tools-version bump, no new dependencies.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,6 +1,12 @@
 // swift-tools-version:6.1
 import PackageDescription
 
+/// `.when(platforms:)` can only include, never exclude, so excluding WASI means listing everything else.
+/// This list matches the [supported platforms on the Swift 6.1 release of SPM](https://github.com/swiftlang/swift-package-manager/blob/release/6.1/Sources/PackageDescription/SupportedPlatforms.swift).
+/// Don't add new platforms here unless raising the swift-tools-version of this manifest.
+let allPlatforms: [Platform] = [.macOS, .macCatalyst, .iOS, .tvOS, .watchOS, .visionOS, .driverKit, .linux, .windows, .android, .wasi, .openbsd]
+let nonWASIPlatforms: [Platform] = allPlatforms.filter { $0 != .wasi }
+
 let package = Package(
     name: "sql-kit",
     platforms: [
@@ -24,7 +30,13 @@ let package = Package(
             dependencies: [
                 .product(name: "Collections", package: "swift-collections"),
                 .product(name: "Logging", package: "swift-log"),
-                .product(name: "NIOCore", package: "swift-nio"),
+                // NIOCore itself builds for wasm32-unknown-wasip1, but the drivers beneath SQLKit
+                // cannot build SwiftNIO there (NIOPosix needs POSIX sockets and threads), so no
+                // driver on that platform could implement an `EventLoopFuture` surface. Target
+                // dependency conditions are evaluated per platform, so on WASI NIOCore is simply
+                // not linked and that surface drops out via `#if canImport(NIOCore)`; the async
+                // surface is unaffected.
+                .product(name: "NIOCore", package: "swift-nio", condition: .when(platforms: nonWASIPlatforms)),
             ],
             swiftSettings: swiftSettings
         ),
@@ -38,8 +50,9 @@ let package = Package(
         .testTarget(
             name: "SQLKitTests",
             dependencies: [
-                .product(name: "NIOCore", package: "swift-nio"),
-                .product(name: "NIOEmbedded", package: "swift-nio"),
+                // The test suite exercises the SwiftNIO surface, so it is not built for WASI.
+                .product(name: "NIOCore", package: "swift-nio", condition: .when(platforms: nonWASIPlatforms)),
+                .product(name: "NIOEmbedded", package: "swift-nio", condition: .when(platforms: nonWASIPlatforms)),
                 .target(name: "SQLKit"),
                 .target(name: "SQLKitBenchmark"),
             ],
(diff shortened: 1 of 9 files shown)


scottmarchant:
Just a heads up, I confirmed that CI checks all pass when ran from the PassiveLogic fork:

https://github.com/PassiveLogic/sql-kit/pull/3/checks

scottmarchant:
I've tested pulling this changes into a private downstream repo of mine and have been able to successfully run sqlite queries running the browser using a Wasm built with these changes.

Package.swift
```diff
@@ -1,6 +1,12 @@
 // swift-tools-version:6.1
 import PackageDescription
 
+/// `.when(platforms:)` can only include, never exclude, so excluding WASI means listing everything else.
+/// This list matches the [supported platforms on the Swift 6.1 release of SPM](https://github.com/swiftlang/swift-package-manager/blob/release/6.1/Sources/PackageDescription/SupportedPlatforms.swift).
+/// Don't add new platforms here unless raising the swift-tools-version of this manifest.
+let allPlatforms: [Platform] = [.macOS, .macCatalyst, .iOS, .tvOS, .watchOS, .visionOS, .driverKit, .linux, .windows, .android, .wasi, .openbsd]
+let nonWASIPlatforms: [Platform] = allPlatforms.filter { $0 != .wasi }
```
MahdiBM, line 8: nitpick: This does not match https://github.com/vapor/sqlite-nio/pull/110/changes#diff-f913940c58e8744a2af1c68b909bb6383e49007e6c5a12fb03104a9006ae677e, let's be consistent.

MahdiBM (commented):
Look good to me but of course Gwynne need to confirm.

Package.swift
MahdiBM: I see the other PR is the one being inconsistent so it's likely easier to just change that PR.

scottmarchant: Great find, thanks! I addressed this in https://github.com/vapor/sqlite-nio/pull/110

PR #203, opened 2026-08-05
Remove SwiftNIO

0xTim:
Remove all the future APIs and the `EventLoop` requirement from the `Database`. This allows us to remove NIO entirely from the package. An alternative to #202, opening for discussion for now as this is obviously a breaking change but could form the basis for Fluent 5. I haven't gone down the full modernisation route of turning on all the flags, removing imports etc as that can be done in future PRs.


scottmarchant:
@0xTim Love this! Just a note, I believe my #202 PR enables conditional nio removal but without requiring a breaking change. Not sure if it makes sense to merge #202 first, then layer #203 on top as the breaking change. If you like that idea, please let me know, I'm happy to look into how the two approaches could work together

